PR #1, opened 2019-03-27
Initial setup

duartepinto:
This PR adds a template iOS objective-c project with travis and codecov configured

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,7 @@ xcode_project: VelocidiSDK/VelocidiSDK.xcodeproj # path to your xcodeproj folder
 xcode_scheme: VelocidiSDKTests
 xcode_destination: platform=iOS Simulator,OS=11.4,name=iPhone X
 
+script: 
+  - xcodebuild -project VelocidiSDK/VelocidiSDK.xcodeproj/ -scheme VelocidiSDKTests clean build test -sdk iphonesimulator -destination 'platform=iOS Simulator,OS=11.4,name=iPhone X'  GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES GCC_GENERATE_TEST_COVERAGE_FILES=YES
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,4 +1,10 @@
 language: objective-c
-xcode_project: VelocidiSDK.xcodeproj # path to your xcodeproj folder
+os: osx
+xcode_project: VelocidiSDK/VelocidiSDK.xcodeproj # path to your xcodeproj folder
 xcode_scheme: VelocidiSDKTests
-xcode_destination: platform=iOS Simulator,OS=10.1,name=iPad Pro (9.7-inch)
+xcode_destination: platform=iOS Simulator,OS=11.4,name=iPhone X
```
antonio-ramadas, line 5: Is it worth to test other iOS versions?

The most recent version is iOS 12 is [installed on 78% of all devices from the last four years (as of January 2019)](https://www.macrumors.com/2019/01/02/ios-12-installed-on-78-percent-of-devices/). The following widest version is iOS 11. Also, while [users on iOS 11 get a chance to update, users on iOS 10 don't](https://everyi.com/by-capability/maximum-supported-ios-version-for-ipod-iphone-ipad.html). The same logic applies to the rest of the iOS versions.

My take is that, depending on the minimum iOS version supported by the SDK, we should test the major versions starting from the minimum. We can take advantage of the [build matrix](https://docs.travis-ci.com/user/build-matrix/) to test the different versions.

What do you think?

antonio-ramadas (changes requested):
From README.md:
> _This is still a very rough draft of the SDK not to ready to be used_

There is a typo. It should be:
_This is still a very rough draft of the SDK not ready to be used_

The message is clear, but what do you think to have instead something like (I'm not sure about the last sentence):
_SDK development in the early stages. Not ready for production. Use it at your own risk._

Side note: I'm writing this here because GitHub does not allow me to write a comment on this line.

duartepinto:
I've run into some problems when setting up Cocoapods and Carthage, and made a lot of changes that will override a lot of the content of this PR. I think it is best to just close this PR and make a single one when I have a more definitive structure

PR #3, opened 2019-04-22
Main structure and tracking

duartepinto:
This PR contains the necessary classes that build the SDK as well as everything necessary to make track requests (model classes and functions).

I've made some of the tests in Swift not only to make them nicer and easier to read but also because it makes it easier to verify if the SDK is nice and usable in Swift (due to how Objective-C is interpreted by Swift). This has caused some problems with travis and forced us to drop some of the jobs that were testing our SDK against multiple versions of the iOS SDK. This happens because in order to have the tests in Swift and the SDK in Objective-C this is dependant on a more recent version of Xcode (10.0) which sets the project as a Legacy Project. We can still test locally against older versions of the iOS SDK, but not with travis because travis bundles each version of the SDK with a version of Xcode.

Not in this PR (we should open issues with these topics):
* Test against redirects. When looking for this, I saw that AFNetwork follows redirects because it is built on top of Foundation URL Loading system which already follows redirects, but testing it would still be a nice to have (https://github.com/AFNetworking/AFNetworking/issues/2387)
* Module and Submodule organization. Currently, we are exposing all classes. It would be nice to research about Submodules and Modules in Objective-C to see if we can set some of the classes (like VSDKUtil) as Internal classes and only expose the classes that we want. Even if we end up exposing everything, it would still be cool to have a proper Module and Submodule organization for better Swift interoperability.
* Divide the Refund model class into TotalRefund and PartialRefund
* Test against multiple SDKs (as mentioned previously)

I know it might be difficult to hop into this PR, because you are not used to working with Objc, so when I was learning I've kept track of the most important things you need to know to understand it.

Coursera course in Objective-C:
* Module 6:
    * Objects
* Module 7:
    * Message Passing 
    * Properties
* Module 10:
    * Callbacks and Blocks

Quick useful links: 
* https://stackoverflow.com/questions/2238584/difference-between-and-methods-in-objective-c
* https://stackoverflow.com/questions/12898820/why-do-we-need-protocol-in-objective-c

More in depth links:
* https://useyourloaf.com/blog/objective-c-class-properties/
* https://medium.com/@MarkEdwardMurray/objective-c-class-initializers-a-k-a-constructors-94dcd6d2ffd7
* https://www.luby.info/2017/09/14/objc-class-property-constants.html
* https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011210-CH1-SW1


antonio-ramadas:
Thank you for pointing out the learning material to review the PR. It definitely helps to get on board 👍

duartepinto:
Regarding the issue with the code coverage dropping to 18%, this is already fixed in another branch. This is caused because Codecov is also taking into account the imported frameworks (which are not being tested).

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductFeedback.h
```diff
@@ -0,0 +1,24 @@
+#import "VSDKTrackingEvent.h"
+
```
antonio-ramadas, line 2: Missing `@class VSDKProduct;`?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTrackingEvent.h
```diff
@@ -0,0 +1,24 @@
+@import JSONModel;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKTrackingEvent : JSONModel
+
+/**
+  Client identifier (REQUIRED)
+ */
+@property (nonnull) NSString *clientId;
```
antonio-ramadas, line 8: Is it required `nonnull`? Shouldn't `NS_ASSUME_NONNULL_BEGIN` be enough to prevent null assignments to variables declared in this region?

Also, can't we also mark the variable as `nonatomic`?

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -0,0 +1,85 @@
...
+class NetworkTests: QuickSpec {
+    override func spec() {
+        super.spec()
+        describe("VelocidiSDK") {
+            context("test track request"){
+                let trackURL = "http://tr.testdomain.com"
+                let matchURL = "http://match.testdomain.com"
+
+                func  responseBuilder(url: String, expectedData: Data) -> (URLRequest) -> (Response) {
+                    return { (request: URLRequest)   in
```
antonio-ramadas:
```suggestion
                    return { (request: URLRequest) in
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
```diff
@@ -0,0 +1,54 @@
+@import JSONModel;
+@class VSDKVoucher;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKTransaction : JSONModel
+
+/**
+  Transactiuon identifier (REQUIRED)
```
antonio-ramadas:
```suggestion
  Transaction identifier (REQUIRED)
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
```diff
@@ -0,0 +1,76 @@
+@import JSONModel;
+
+NS_ASSUME_NONNULL_BEGIN
```
antonio-ramadas: Shouldn't this macro be removed? All properties except one are optional (nullable), right?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
```diff
@@ -0,0 +1,76 @@
...
+
+@interface VSDKProduct : JSONModel
+
+/**
+  Product identifier (REQUIRED)
```
antonio-ramadas: I keep seeing this _REQUIRED_ spread across the files, but is it really necessary? Doesn't `nonnull` already indicate this?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
```diff
@@ -0,0 +1,54 @@
...
+ */ @property (nonnull) NSString *transactionId;
```
antonio-ramadas: Missing a new line

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+  Transaction total price (after taxes and shipping)
+ */
+@property float price;
+
+/**
+  Transaction frequency using cron format (https://crontab.guru/)
+ */
+@property NSString *recurrence;
```
antonio-ramadas: Are you sure this property exists?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
```diff
@@ -0,0 +1,54 @@
+@import JSONModel;
+@class VSDKVoucher;
+
+NS_ASSUME_NONNULL_BEGIN
```
antonio-ramadas: Same as the comment I left at `VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h`.

P.S.: I could have left here a link to the comment I mentioned, but while I don't finish the review I am not able to copy it.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKAddToCart.h
```diff
@@ -0,0 +1,16 @@
+#import "VSDKTrackingEvent.h"
+
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKAddToCart : VSDKTrackingEvent
+
+/**
+  Product added to the cart
```
antonio-ramadas:
```suggestion
  Products added to the cart
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKAddToCart.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *product;
```
antonio-ramadas:
```suggestion
@property NSMutableArray<VSDKProduct *> *products;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKPageView.h
```diff
@@ -0,0 +1,29 @@
+#import "VSDKTrackingEvent.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKPageView : VSDKTrackingEvent
+
+/**
+  Page being viewed
```
antonio-ramadas: What do you think about having `Address of the page being seen` instead?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKCustomization.h
```diff
@@ -0,0 +1,29 @@
...
+
+@interface VSDKCustomization : JSONModel
+
+/**
+  Product customization name
+ */
+@property NSString *name;
+
+/**
+  Product customization
```
antonio-ramadas: I believe `Custom property to add extra information about the customization` sums it better. What do you think?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductClick.h
```diff
@@ -0,0 +1,15 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKProductClick : VSDKTrackingEvent
+
+/**
+  Product associated with this product click
```
antonio-ramadas:
```suggestion
  The clicked products
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductClick.h
```diff
@@ -0,0 +1,15 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *product;
```
antonio-ramadas:
```suggestion
@property NSMutableArray<VSDKProduct *> *products;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
```diff
@@ -0,0 +1,21 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+@class VSDKCustomization;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKProductCustomization : VSDKTrackingEvent
+
+/**
+  Product associated with this product customization
```
antonio-ramadas:
```suggestion
  Products being customized
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *product;
```
antonio-ramadas:
```suggestion
@property NSMutableArray<VSDKProduct *> *products;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+  Customization associated with this product customization
```
antonio-ramadas: Customization being applied to the products

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+@property VSDKCustomization *productCustomization;
```
antonio-ramadas:
```suggestion
@property VSDKCustomization *customization;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductView.h
```diff
@@ -0,0 +1,15 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKProductView : VSDKTrackingEvent
+
+/**
+  List of products associated with this product view
```
antonio-ramadas:
```suggestion
  List of products seen
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductViewDetails.h
```diff
@@ -0,0 +1,15 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKProductViewDetails : VSDKTrackingEvent
+
+/**
+  List of products associated with this product view details
```
antonio-ramadas:
```suggestion
  List of products whose details have been seen
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKPurchase.h
```diff
@@ -0,0 +1,15 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKPurchase : VSDKTrackingEvent
+
+/**
+  Product purchased
```
antonio-ramadas:
```suggestion
  Products purchased
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKPurchase.h
```diff
@@ -0,0 +1,15 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *product;
```
antonio-ramadas:
```suggestion
@property NSMutableArray<VSDKProduct *> *products;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
```diff
@@ -0,0 +1,28 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKTransaction;
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+//TODO Split this into VSDKPartialRefund and VSDKTotalRefund
```
antonio-ramadas, line 5: Is it to solve in a future PR?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
```diff
@@ -0,0 +1,28 @@
...
+@interface VSDKRefund : VSDKTrackingEvent
+
+/** 
+  Type of refund 
+  A refund can be either `partial` or `total`
+ */
+@property NSString *refundType;
+
+/**
+  If it is a `partial` refund, the list of products to be refunded must be specified
```
antonio-ramadas, line 15: Can we enforce this logic in the constructor and setters? Perhaps `NSAssert` can help us here.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRemoveFromCart.h
```diff
@@ -0,0 +1,15 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKRemoveFromCart : VSDKTrackingEvent
+
+/**
+  Product removed form the cart
```
antonio-ramadas:
```suggestion
  Products removed from the cart
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRemoveFromCart.h
```diff
@@ -0,0 +1,15 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *product;
```
antonio-ramadas:
```suggestion
@property NSMutableArray<VSDKProduct *> *products;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKSearch.h
```diff
@@ -0,0 +1,14 @@
+#import "VSDKTrackingEvent.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKSearch : VSDKTrackingEvent
+
+/**
+  Query searched by the user
```
antonio-ramadas, line 6:
```suggestion
  Search query used by the user
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKSubscription.h
```diff
@@ -0,0 +1,21 @@
+#import "VSDKTrackingEvent.h"
+@class VSDKTransaction;
+@class VSDKProduct;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKSubscription : VSDKTrackingEvent
+
+/**
+  Products associated with to the Subscription
```
antonio-ramadas:
```suggestion
  Products associated with the Subscription
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKSubscription.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+@property NSMutableArray<VSDKProduct *> *products;
+
+/**
+  Transaction associated with to the Subscription
```
antonio-ramadas:
```suggestion
  Transaction associated with the Subscription
```

VelocidiSDK/VelocidiSDK/VelocidiSDK.h
```diff
@@ -8,4 +8,22 @@ FOUNDATION_EXPORT const unsigned char VelocidiSDKVersionString[];
...
+#import "VSDKProduct.h"
+#import "VSDKProductView.h"
+#import "VSDKProductCustomization.h"
+#import "VSDKAddToCart.h"
+#import "VSDKProductViewDetails.h"
+#import "VSDKTransaction.h"
+#import "VSDKVelocidi.h"
+#import "VSDKConfig.h"
+#import "VSDKRequest.h"
+#import "VSDKUtil.h"
```
antonio-ramadas, line 28: Can you sort the imports?

```objc
#import "VSDKAddToCart.h"
#import "VSDKConfig.h"
#import "VSDKCustomization.h"
#import "VSDKPageView.h"
#import "VSDKProduct.h"
#import "VSDKProductClick.h"
#import "VSDKProductCustomization.h"
#import "VSDKProductFeedback.h"
#import "VSDKProductImpression.h"
#import "VSDKProductView.h"
#import "VSDKProductViewDetails.h"
#import "VSDKPurchase.h"
#import "VSDKRefund.h"
#import "VSDKRequest.h"
#import "VSDKSearch.h"
#import "VSDKTransaction.h"
#import "VSDKUtil.h"
#import "VSDKVelocidi.h"
#import "VSDKVoucher.h"
```

VelocidiSDK/VelocidiSDK/VelocidiSDK.h
```diff
@@ -8,4 +8,22 @@ FOUNDATION_EXPORT const unsigned char VelocidiSDKVersionString[];
 
 // In this header, you should import all the public headers of your framework using statements like #import <VelocidiSDK/PublicHeader.h>
```
antonio-ramadas:
```suggestion
```

VelocidiSDK/VelocidiSDK/VSDKVelocidi.h
```diff
@@ -0,0 +1,54 @@
+#import <UIKit/UIKit.h>
+@class VSDKConfig;
+@class VSDKTrackingEvent;
+@class AFHTTPSessionManager;
+
+NS_ASSUME_NONNULL_BEGIN
+
```
antonio-ramadas, line 7: Nitpick:

```suggestion
```

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,24 @@
+#import <Foundation/Foundation.h>
+
```
antonio-ramadas, line 2: Nitpick: 
```suggestion
```

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,24 @@
...
+
+@interface VSDKConfig : NSObject
+
+/**
+  URL used to make track requests
+ */
+@property NSString *trackingHost;
```
antonio-ramadas: Wouldn't it be better to use [`NSURL`](https://developer.apple.com/documentation/foundation/nsurl) instead?

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+  URL used to make match requests
+ */
+@property NSString *matchHost;
```
antonio-ramadas: Same as `trackingHost`, wouldn't it be better to use [`NSURL`](https://developer.apple.com/documentation/foundation/nsurl) instead?

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,24 @@
+#import <Foundation/Foundation.h>
+
+
+@interface VSDKConfig : NSObject
+
+/**
+  URL used to make track requests
```
antonio-ramadas:
```suggestion
  URL endpoint to make track requests
```

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+@property NSString *trackingHost;
+
+/**
+  URL used to make match requests
```
antonio-ramadas:
```suggestion
  URL endpoint to make match requests
```

VelocidiSDK/VelocidiSDK/VSDKVelocidi.h
```diff
@@ -0,0 +1,54 @@
...
+ */
++ (instancetype)sharedInstance;
+
+/**
+  Method that takes the provided configuration and properly initializes the Velocidi SDK according 
+  to the settings of said configuration.
+  @param config necessary configuration to properly start the VSDKVelocidi
+  @returns shared single instance of the class VSDKVelocidi
+ */
++ (instancetype) start: (VSDKConfig *)config;
```
antonio-ramadas:
```suggestion
+ (instancetype)start: (VSDKConfig *)config;
```

VelocidiSDK/VelocidiSDK/VSDKUtil.m
```diff
@@ -0,0 +1,63 @@
...
+    return userAgent;
+    
+}
+
+- (NSUUID *)getAdvertisingIdentifier {
+    return [[ASIdentifierManager sharedManager] advertisingIdentifier];
+}
+
+- (bool)isTrackAllowed:(NSUUID *)advertisingIdentifier {
+    if ([[advertisingIdentifier UUIDString] isEqualToString:@"00000000-0000-0000-0000-000000000000"]){
```
antonio-ramadas: Why not use [`isAdvertisingTrackingEnabled `](https://developer.apple.com/documentation/adsupport/asidentifiermanager/1614148-isadvertisingtrackingenabled)?

antonio-ramadas (changes requested):
I'm still reviewing, but I'm going to wrap up for today. Tomorrow I'll continue.

I'm a bit uncertain about the spread usage of `NS_ASSUME_NONNULL_BEGIN`. Some comments reflect that. I noticed some other places, but for the sake of avoiding repeating myself, I stopped adding comments for this case and see where the discussion ends. In case we decide to change it, I don't mind adding comments in the rest of the places where I detected it.

VelocidiSDK/VelocidiSDK/VSDKUtil.h
```diff
@@ -0,0 +1,36 @@
...
+- (NSString *)getVersionedUserAgent;
+
+/**
+  Retrieve the advertisingIdentifier UUID.
+  @returns advertisingIdentifier UUID.
+ */
+- (NSUUID *)getAdvertisingIdentifier;
+
+/**
+  Verify if tracking is allowed or if the user has opted-out of being tracked in this application.
```
antonio-ramadas: `if tracking is allowed` and `if the user has opted-out of being tracked in this application` aren't these the same?

VelocidiSDK/VelocidiSDK/VSDKUtil.m
antonio-ramadas: Forget it. I've only seen the tests now. You use this method to mock this class.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
```diff
@@ -0,0 +1,63 @@
+#import "VSDKUtil.h"
+#import <AdSupport/ASIdentifierManager.h>
+
+@implementation VSDKUtil
+
+- (NSString *)getVersionedUserAgent {
```
antonio-ramadas, line 5: I noticed some duplicated code (the user agent on all devices shares the same prefix). What if we move this prefix construction to the top and only append the rest of the string?

```diff
diff --git a/VelocidiSDK/VelocidiSDK/VSDKUtil.m b/VelocidiSDK/VelocidiSDK/VSDKUtil.m
index 2ffecad..4b460bb 100644
--- a/VelocidiSDK/VelocidiSDK/VSDKUtil.m
+++ b/VelocidiSDK/VelocidiSDK/VSDKUtil.m
@@ -4,33 +4,25 @@
 @implementation VSDKUtil
 
 - (NSString *)getVersionedUserAgent {
+    NSString *userAgentPrefix = [NSString stringWithFormat:@"%@/%@ %@/%@",
+                                 [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleExecutableKey] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleIdentifierKey],
+                                 [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleVersionKey],
+                                 @"VelocidiSDK",
+                                 @(VelocidiSDKVersionNumber)];
     NSString *userAgent = nil;
+    
 #if TARGET_OS_IOS
-    // User-Agent Header; see http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.43
-    userAgent = [NSString stringWithFormat:@"%@/%@ %@/%@ (%@; iOS %@; Scale/%0.2f)",
-                 [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleExecutableKey] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleIdentifierKey],
-                 [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleVersionKey],
-                 @"VelocidiSDK",
-                 @(VelocidiSDKVersionNumber),
+    userAgent = [userAgentPrefix stringByAppendingFormat:@" (%@; iOS %@; Scale/%0.2f)",
                  [[UIDevice currentDevice] model],
                  [[UIDevice currentDevice] systemVersion],
                  [[UIScreen mainScreen] scale]];
 #elif TARGET_OS_WATCH
-    // User-Agent Header; see http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.43
-    userAgent = [NSString stringWithFormat:@"%@/%@ %@/%@ (%@; watchOS %@; Scale/%0.2f)",
-                 [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleExecutableKey] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleIdentifierKey],
-                 [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleVersionKey],
-                 @"VelocidiSDK",
-                 @(VelocidiSDKVersionNumber),
+    userAgent = [userAgentPrefix stringByAppendingFormat:@" (%@; watchOS %@; Scale/%0.2f)",
                  [[WKInterfaceDevice currentDevice] model],
                  [[WKInterfaceDevice currentDevice] systemVersion],
                  [[WKInterfaceDevice currentDevice] screenScale]];
 #elif defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
-    userAgent = [NSString stringWithFormat:@"%@/%@ %@/%@ (Mac OS X %@)",
-                 [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleExecutableKey] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleIdentifierKey],
-                 [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] ?: [[NSBundle mainBundle] infoDictionary][(__bridge NSString *)kCFBundleVersionKey],
-                 @"VelocidiSDK",
-                 @(VelocidiSDKVersionNumber),
+    userAgent = [userAgentPrefix stringByAppendingFormat:@" (Mac OS X %@)",
                  [[NSProcessInfo processInfo] operatingSystemVersionString]];
 #endif
     if (userAgent) {
@@ -42,7 +34,6 @@ - (NSString *)getVersionedUserAgent {
         }
     }
     return userAgent;
-    
 }
 
 - (NSUUID *)getAdvertisingIdentifier {

```

Also, the user agent comment would be better in the header file, right? Something along the lines of: `User-Agent Header follows the guidelines of RFC 2616 (http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.43 )`. What do you think?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTrackingEvent.h
duartepinto: > Is it required nonnull? Shouldn't NS_ASSUME_NONNULL_BEGIN be enough to prevent null assignments to variables declared in this region?

No, it is not required, but I prefer to remove the `NS_ASSUME_NONNULL_BEGIN` in all model classes (to make it more explicit to the developer which fields are optional and which fields aren't)

> Also, can't we also mark the variable as nonatomic?

 What benefit do you see in that?

duartepinto:
> I'm a bit uncertain about the spread usage of `NS_ASSUME_NONNULL_BEGIN`. Some comments reflect that. I noticed some other places, but for the sake of avoiding repeating myself, I stopped adding comments for this case and see where the discussion ends. In case we decide to change it, I don't mind adding comments in the rest of the places where I detected it.

Yes, I agree we can remove the `NS_ASSUME_NONNULL_BEGIN/END` in all tracking model classes

VelocidiSDK/VelocidiSDK/VSDKRequest.h
```diff
@@ -0,0 +1,41 @@
...
+        :(void (^)(NSError *error))onFailureBlock;
+
+/**
+  Build the parameters portion of the request URL.
+  The built string contains the `cookies` parameter set to false, and the `id_idfa` parameter with the
+  advertisingIdentifier of the user.
+  @param advertisingIdentifier advertisingIdentifier of the user
+  @returns URL parameters string
+ */
+- (NSString *)buildURLParameters:(NSString *) advertisingIdentifier;
```
antonio-ramadas: Wouldn't it be better if we returned instead [`NSURLComponents`](https://developer.apple.com/documentation/foundation/nsurlcomponents) or an array of [`NSURLQueryItem`](https://developer.apple.com/documentation/foundation/nsurlqueryitem)?

VelocidiSDK/VelocidiSDK/VSDKRequest.m
```diff
@@ -0,0 +1,60 @@
+@import Foundation;
+#import "VSDKRequest.h"
+#import "VSDKUtil.h"
+
+@implementation VSDKRequest
+
+- (instancetype)initWithHTTPSessionManager:(AFHTTPSessionManager *)manager{
+
```
antonio-ramadas, line 20:
```suggestion
```

VelocidiSDK/VelocidiSDK/VSDKRequest.m
```diff
@@ -0,0 +1,60 @@
...
+    NSParameterAssert(manager != nil);
```
antonio-ramadas: The header file has `NS_ASSUME_NONNULL_BEGIN`. So, this is not really necessary, right?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
duartepinto: I know it is a bit redundant, but I would prefer to leave it there so that it's really clear that said property is _REQUIRED_. But that is a really personal preference if you want I can remove it 😅.

VelocidiSDK/VelocidiSDK/VSDKRequest.h
```diff
@@ -0,0 +1,41 @@
...
+
+@class VSDKUtil;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface VSDKRequest<__covariant T:JSONModel *> : NSObject
+
+@property (readonly, nonatomic) AFHTTPSessionManager *manager;
+@property NSURL *url;
+@property (nonatomic) VSDKUtil *util;
```
antonio-ramadas:
```suggestion
@property (readonly, nonatomic) VSDKUtil *util;
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
duartepinto: Removed in 0f0ef14

VelocidiSDK/VelocidiSDK/VSDKConfig.h
duartepinto: Yes, I agree. Changed it in 47b78a1

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: ~Tracking is more something that the user allows or disallows, not soo much as something that is enabled or disabled 😅`.~

VelocidiSDK/VelocidiSDKTests/SanityTests.m
```diff
@@ -0,0 +1,36 @@
...
+
+@implementation SanityTests
+
+- (void)testJsonModel {
+
+    VSDKTrackingEvent * trackingEvent = [[VSDKPageView alloc] init];
+    trackingEvent.clientId = @"0";
+    trackingEvent.siteId = @"1";
+    XCTAssert([@"{\"siteId\":\"1\",\"clientId\":\"0\",\"type\":\"pageView\"}"
+            isEqualToString:[trackingEvent toJSONString]]
```
antonio-ramadas, line 15: Can we make this comparison as JSON?

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: Nvm, I was not aware of https://developer.apple.com/documentation/adsupport/asidentifiermanager/1614148-isadvertisingtrackingenabled. I thought that you were suggesting renaming that function. Using a native function is indeed much better, I totally missed that one.

VelocidiSDK/VelocidiSDKTests/SanityTests.m
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// FIXME This test has an 'A' in the beginning because xcode runs the tests in alphabetical order
+- (void)AtestVelocidiInstanceFail {
+    XCTAssertThrowsSpecificNamed(
+                                 VSDKVelocidi.sharedInstance,
+                                 NSException,
+                                 NSInternalInconsistencyException
+                                 );
+
```
antonio-ramadas:
```suggestion
```

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
```diff
@@ -0,0 +1,3 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
+//
```
antonio-ramadas: Is this file necessary?

VelocidiSDK/VelocidiSDKTests/UtilTests.swift
```diff
@@ -0,0 +1,25 @@
...
+        super.spec()
+        describe("VSDKUtil") {
+            it("should generate a valid User-Agent containing information about VelocidiSDK") {
+                let userAgent = VSDKUtil().getVersionedUserAgent()
+                let regexStr = String(format:".*/.* VelocidiSDK\\/%0.0f \\(%@; %@ %@; Scale\\/%0.2f\\)",
+                    VelocidiSDKVersionNumber,
+                    UIDevice.current.model,
+                    UIDevice.current.systemName,
+                    UIDevice.current.systemVersion,
+                    UIScreen.main.scale
```
antonio-ramadas:
```suggestion
                    UIScreen.main.scale)
```

VelocidiSDK/VelocidiSDKTests/UtilTests.swift
```diff
@@ -0,0 +1,25 @@
...
+                     )
```
antonio-ramadas:
```suggestion
```

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: Changed it in 81466fd. I still left the function in VSDKUtil, so that it is easier to mock and test it. Let me know what you think about that. I know that it is a bit silly to have all that just to test an `if` but I still think it would be good to have a test that verifies that we are not tracking when tracking is not allowed.

VelocidiSDK/VelocidiSDKTests/RequestsTests.swift
```diff
@@ -0,0 +1,53 @@
...
+                let url = "http://testdomain.com"
+                let trackingEvent = VSDKPageView()
+                trackingEvent.siteId = "0"
+                trackingEvent.clientId = "0"
+                
+                var requestExecuted: Bool = false
+
+                self.stub({(request: URLRequest) in
+                    return request.url!.absoluteString.starts(with: url)
+                }, failure(NSError(domain: url, code: 451)))
```
antonio-ramadas: What is the meaning of 451? Is the HTTP error code? Wouldn't it be better a [409](https://httpstatuses.com/409)?

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -0,0 +1,85 @@
...
+                let trackURL = "http://tr.testdomain.com"
+                let matchURL = "http://match.testdomain.com"
+
+                func  responseBuilder(url: String, expectedData: Data) -> (URLRequest) -> (Response) {
+                    return { (request: URLRequest) in
+                        if(request.url!.absoluteString.starts(with: url)
+                            && request.allHTTPHeaderFields?["Content-Type"] == "application/json"
+                            && request.allHTTPHeaderFields?["User-Agent"] != nil
+                            && request.httpBodyStream != nil
+                            ){
```
antonio-ramadas:
```suggestion
```

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -0,0 +1,85 @@
...
+            context("test track request"){
+                let trackURL = "http://tr.testdomain.com"
+                let matchURL = "http://match.testdomain.com"
+
+                func  responseBuilder(url: String, expectedData: Data) -> (URLRequest) -> (Response) {
+                    return { (request: URLRequest) in
+                        if(request.url!.absoluteString.starts(with: url)
+                            && request.allHTTPHeaderFields?["Content-Type"] == "application/json"
+                            && request.allHTTPHeaderFields?["User-Agent"] != nil
+                            && request.httpBodyStream != nil
```
antonio-ramadas:
```suggestion
                            && request.httpBodyStream != nil) {
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductFeedback.h
duartepinto: Yes, fixed it already in 0f0ef14

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -0,0 +1,85 @@
...
+
+                    var success: Bool? = nil
+
+                    let config = VSDKConfig(hosts:trackURL, matchURL)
+                    VSDKVelocidi.start(config!)
+                    
+                    VSDKVelocidi.sharedInstance().track(trackingEvent, onSuccess:{ (response: URLResponse, responseObject: Any) in
+                        success = true
+                    }, onFailure:{(error: Error) in
+                        NSLog("Error \(error.localizedDescription)")
```
antonio-ramadas, line 72: https://stackoverflow.com/a/19201849

```suggestion
                        NSLog(@"Error %@", error.localizedDescription)
```

antonio-ramadas (changes requested):
Continuation of the last review.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
duartepinto: This is not possible because the backend is expecting a `productCustomization` (see https://docs.velocidi.com/knowledgebase/tracking-api/#productCustomization)

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
duartepinto: Not in any of the pending PRs but I preferred to leave it like this for now and to iterate on this issue in the future.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
duartepinto: Because we will probably redo this, I prefer to leave this change for a future PR.

VelocidiSDK/VelocidiSDK/VSDKConfig.h
duartepinto: Fixed in 47b78a1

VelocidiSDK/VelocidiSDK/VelocidiSDK.h
duartepinto: Changed in 0693481

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKSearch.h
duartepinto: "Search query used" sounds strange 😅

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
duartepinto: I believe so, check https://docs.velocidi.com/knowledgebase/tracking-api/#subscriptions

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKPageView.h
duartepinto: Changed in 2b9650f

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKCustomization.h
duartepinto: Changed in 2b9650f

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
duartepinto: Added in 2b9650f

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductCustomization.h
duartepinto: Changed in 2b9650f

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductView.h
duartepinto: Changed in 2b9650f but used the verb _view_ instead of _see_ for consistency

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProductViewDetails.h
duartepinto: Changed in 2b9650f but used the verb _view_ instead of _see_ for consistency

VelocidiSDK/VelocidiSDKTests/RequestsTests.swift
duartepinto: There isn't a specific motive, just thought it would be the most appropriate from all the 4XX error codes. 409 also does not seem like the best option (_conflict with the current state of the **target** resource_ is not the case here). Perhaps 403 is the best option!

VelocidiSDK/VelocidiSDK/VSDKUtil.h
duartepinto: The _or_ here is exclusive.

VelocidiSDK/VelocidiSDK/VSDKRequest.m
duartepinto: Good catch! Just tried it and the compiler gives a warning when using a null. Removed in 31d2e81

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
duartepinto: That's for Objective-C. In Swift, there is no need to use format specifiers.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: Took most of this implementation from https://github.com/AFNetworking/AFNetworking (I've added a mention to that in another branch that will come in a future PR) and I would prefer to leave the implementation as similar to the original as possible (to make it easier to `diff`).

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
duartepinto: Yes. Tried to remove it and had some problems when executing the tests in Travis

VelocidiSDK/VelocidiSDKTests/RequestsTests.swift
antonio-ramadas: Sure, 403 seems fine 🙂

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTrackingEvent.h
antonio-ramadas: > No, it is not required, but I prefer to remove the `NS_ASSUME_NONNULL_BEGIN` in all model classes (to make it more explicit to the developer which fields are optional and which fields aren't)

👍 

> > Also, can't we also mark the variable as nonatomic?
> 
> What benefit do you see in that?

[Avoid thread safety boilerplate](https://stackoverflow.com/a/589348)

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
antonio-ramadas: Let's see what @jcazevedo has to say

VelocidiSDK/VelocidiSDK/VSDKRequest.h
duartepinto: Implemented this but later reverted it in cd74998. Doing this makes testing difficult, as a mock VSDKRequest cannot assign `util`

VelocidiSDK/VelocidiSDK/VSDKRequest.h
duartepinto: Yes, that is indeed much better. Added it in cd74998

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
antonio-ramadas: Can you create an issue and reference each other on both sides?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
antonio-ramadas: Are you referring to the TODO above? Unless there is a concrete task for us to do it in the future, I don't know when it will be done. Also, that TODO isn't a restriction for us to ship the SDK, right? Why not implement this logic right away? We will be more relaxed to ship this release and the user will have more trouble to create an invalid event.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
antonio-ramadas: I think we can improve this a little bit.

Why don't we have a private property named `sharedManager` of type `ASIdentifierManager`? When using it, if it is `nil`, then we set it to `[ASIdentifierManager sharedManager]`. This allows us to create a mock class which implements `ASIdentifierManager` and we set it to `sharedManager` (the new property we created).

I haven't tested this, but it should work. I'm only worried if implementing/extending `ASIdentifierManager` is possible.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
antonio-ramadas: Can you comment here some direct links to it?

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
antonio-ramadas: Hum... That seems weird. Can you explain why or paste here the link of the faulty Travis? I'd like to know more.

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
antonio-ramadas: Are you sure? https://developer.apple.com/documentation/foundation/1409759-nslog#

VelocidiSDK/VelocidiSDK/VSDKUtil.h
antonio-ramadas: Let me rephrase, tracking is only allowed if the user has NOT opted-out of being tracked in this application. It seems to me the first condition is a derivative of the second one, right?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
duartepinto: > Are you referring to the TODO above?

Yes. 

> Unless there is a concrete task for us to do it in the future, I don't know when it will be done. Also, that TODO isn't a restriction for us to ship the SDK, right? Why not implement this logic right away? We will be more relaxed to ship this release and the user will have more trouble to create an invalid event.

I prefer to leave the model classes a bit crude because they are not the main focus of this PR and we can iterate on that later and add other quality of life improvements, like proper initializers (to avoid using the default `init` and enforce the initialization of the required properties). As it is, this PR is already really big and I would prefer to have changes that add functionality and can be packed to come in a future PR.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h
duartepinto: After discussing this offline, we came to the realization that this property does not exist. Removed it in 9579f1f

VelocidiSDK/VelocidiSDKTests/RequestsTests.swift
duartepinto: Changed in 9579f1f

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: https://github.com/AFNetworking/AFNetworking/blob/009e3bb6673edc183c4f2baf552ad7cccba94d58/AFNetworking/AFURLRequestSerialization.m#L221

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
duartepinto: Swift still allows the use of format specifiers, but it is not needed and we can use string interpolation if it is easier.

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
duartepinto: Removing this file in Xcode results in the following error when trying to run the tests (not only in Travis): 

```
2019-05-07 18:35:55.804 xcodebuild[51558:891558]  DVTAssertions: Warning in /Library/Caches/com.apple.xbs/Sources/IDEFrameworks_Fall2018/IDEFrameworks-14460.46/IDEFoundation/Logging/ActivityLog/IDEActivityLogSectionRecorder.m:1050
Details:  log recorder was sent -stopRecordingWithInfo:completionBlock: after it had already been asked to stop recording.
Object:   <IDEActivityLogSectionRecorder: 0x7f8b43f360e0>
Method:   -stopRecordingWithInfo:completionBlock:
Thread:   <NSThread: 0x7f8b4389a900>{number = 15, name = (null)}
Please file a bug at https://bugreport.apple.com with this warning message and any useful information you can provide.

Testing failed:
        Error opening input file '/Users/duarte/Documents/git/velocidi-ios-objc-sdk/VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h' (No such file or directory)
        Testing cancelled because the build failed.
** TEST FAILED **


The following build commands failed:
        PrecompileSwiftBridgingHeader normal x86_64
        CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(2 failures)
```

Removed this using Xcode so it should remove all references to this file but somewhere in the configurations it is still pointing to it. Even if it is empty, it might be needed because of interoperability and because we have tests in both Objective-C and Swift. But this is just out of the top of my head, I have not researched this, I'll look into it and come back later

VelocidiSDK/VelocidiSDK/VSDKUtil.m
antonio-ramadas: I have to say I prefer the refactoring I posted. It is much easier to read and understand.

antonio-ramadas:
There is one other thing I almost forgot, every time the user declares an event, he has to set the `siteId` and `clientId`, right? Given that they will be most likely constant, can we make the API possible to set them only once and propagate them to every event generated? For instance, these variables by default could be read from the environment variables or even be passed as command line arguments.

antonio-ramadas:
> Yes, I agree we can remove the `NS_ASSUME_NONNULL_BEGIN/END` in all tracking model classes

Xcode shows a bunch of warnings. Can you fix them? I presume it was caused by the removal of this macro.

duartepinto:
> Xcode shows a bunch of warnings. Can you fix them? I presume it was caused by the removal of this macro.

Can you post some screenshots or logs of the errors? Could not replicate this.

duartepinto:
> There is one other thing I almost forgot, every time the user declares an event, he has to set the `siteId` and `clientId`, right? Given that they will be most likely constant, can we make the API possible to set them only once and propagate them to every event generated? For instance, these variables by default could be read from the environment variables or even be passed as command line arguments.

Do you have any suggestion about how we would implement this? This would be nice, but if we were to do this I would like it to be agnostic of any logic (making it for any values, not only for the `siteId` or `clientId`), something like an auto-filler that would populate instances of `VSDKTrackingEvent` with a set of predefined values selected by the developer. Something like this:

```objc
NSDictionary * predefinedValues =  @{@"siteId" : @"1", @"clientId" : @"2", @"pageType" : @"fooPage" };

VSDKTrackingEvent * event = [[VSDKPageView alloc] init];
event.title = @"fooTitle";

[VSDKTrackingEventFiller config: predefinedValues event: event]; //Merge the event with the predefinedValues
```

But I think doing something like this would be quite complex and don't know if we should bundle it in this PR.

VelocidiSDK/VelocidiSDKTests/SanityTests.m
duartepinto: Don't think so. There is no JSON like class in the JSONModel, just converts from a source (like a String) that can be interpreted as a JSON and models it to a class

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
duartepinto: Yes, that file is needed. The VelocidiSDKTests target is a mixed Objective-C/Swift target and it needs a Bridging header for interoperability. Although we don't have any interoperability between different test classes, we could. More: https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/importing_objective-c_into_swift

antonio-ramadas:
> > Xcode shows a bunch of warnings. Can you fix them? I presume it was caused by the removal of this macro.
> 
> Can you post some screenshots or logs of the errors? Could not replicate this.

Here is one:
```
velocidi-ios-objc-sdk/VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTransaction.h:19:20: Pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified)
```

Screenshot 2019-05-09 at 11 50 30

VelocidiSDK/VelocidiSDKTests/VelocidiSDKTests-Bridging-Header.h
antonio-ramadas: I believe I managed to do it. The build succeeds and the tests pass! Move the file to trash and [follow these steps](https://stackoverflow.com/a/32274718).

duartepinto: Nice! Removed it in 9cd4b6e

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: What do you think @jcazevedo?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKRefund.h
duartepinto: I'll create it after this PR is merged (just for consistency)

duartepinto:
> There is one other thing I almost forgot, every time the user declares an event, he has to set the `siteId` and `clientId`, right? Given that they will be most likely constant, can we make the API possible to set them only once and propagate them to every event generated? For instance, these variables by default could be read from the environment variables or even be passed as command line arguments.

Implemented in e4a3f01

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKTrackingEvent.h
duartepinto: During an offline conversation, we decided not to do this

VelocidiSDK/VelocidiSDK/VSDKUtil.h
duartepinto: Changed it in fb202aa

VelocidiSDK/VelocidiSDK/VSDKUtil.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+  Retrieve the advertisingIdentifier UUID.
+  @returns advertisingIdentifier UUID.
+ */
+- (NSUUID *)getAdvertisingIdentifier;
+
+/**
+  Verify if tracking is allowed.
+  Tracking is not allowed if user has opted-out of ad tracking (Limited Ad Tracking is enabled in the user's device).
```
antonio-ramadas:
```suggestion
  Tracking is not allowed if the user has opted-out of ad tracking (Limited Ad Tracking is enabled in the user's device).
```

VelocidiSDK/VelocidiSDK/VSDKConfig.h
```diff
@@ -0,0 +1,23 @@
+#import <Foundation/Foundation.h>
+
+@interface VSDKConfig : NSObject
+
+/**
+  URL endpoint to make track requests
+ */
+@property NSURL *trackingHost;
```
jcazevedo, line 8: We're capable of deriving the endpoints for our HTTP requests given the subdomain of the target system. I think this configuration can be simplified if we simply require that.

VelocidiSDK.xcworkspace/xcshareddata/IDETemplateMacros.plist
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>FILEHEADER</key>
+    <string> Created for ___PROJECTNAME___ in ___YEAR___
+// Copyright (c) ___YEAR___ ___ORGANIZATIONNAME___. 
+// Licensed under the Apache License, Version 2.0 (the "License");
+    </string>
```
jcazevedo, line 9: Since we don't seem to be using this header anywhere, I'd remove the value of the `FILEHEADER` key. Unfortunately, this still adds a line with `//` to new files, but there's an open issue to address that: https://openradar.appspot.com/41383331.

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
```diff
@@ -0,0 +1,72 @@
...
+ */
+@property NSString *brand;
+
+/**
+  Product category
+ */
+@property NSString *category;
+
+/**
+  Product variante (e.g. color, size)
```
jcazevedo:
```suggestion
  Product variant (e.g. color, size)
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
jcazevedo: > I know it is a bit redundant, but I would prefer to leave it there so that it's really clear that said property is *REQUIRED*.

In the absence of the possibility of making this required using the type system, I think a compiler warning is a more clear indication than a code comment. Having that, I think the comment is unnecessary. 

Do we also need comments in all these properties? Some are just a copy of the name of the property.

RELEASING.md
```diff
@@ -4,14 +4,36 @@ _This is still a draft of the process, it needs verification and a test run in a
...
+    `$ cd VelocidiSDK`
+
+1. Edit the version number and increment the build number using agvtool (more info [here](https://developer.apple.com/library/archive/qa/qa1827/_index.html)):
+    1. __Version number__:
+
+        `$ agvtool new-marketing-version <your_specific_version>`
+
+    1. __Build number__:
+    
+        `$ agvtool new-version -all`
```
jcazevedo: I think this is lacking a version number.

```suggestion
        `$ agvtool new-version -all <your_specific_version>`
```

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+  Quantity of the product (REQUIRED)
+  Default is 1
+ */
+@property int quantity;
+
+/**
+  Whether or not the product was suggested by a recommendation.
+ */
+@property NSString *recomendation;
```
jcazevedo: Shouldn't this be a boolean?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+  Whether or not the product is unsafe (e.g. adult content)
+ */
+@property NSString *unsafe;
```
jcazevedo: Shouldn't this be a boolean?

VelocidiSDK/VelocidiSDK/VSDKUtil.m
jcazevedo: I don't have a strong opinion one way or the other, but since this is isn't an ipsis verbis copy of the original source, I don't think the suggested change makes the diff that much harder to parse.

VelocidiSDK/VelocidiSDK/VSDKRequest.m
```diff
@@ -0,0 +1,63 @@
...
+        _util = [[VSDKUtil alloc] init] ;
+        [VSDKUtil setAcceptAllResponses:_manager];
+    }
+    return self;
+}
+
+- (void)performRequest: (void (^)(NSURLResponse *response, id responseObject))onSuccessBlock
+        :(void (^)(NSError *error))onFailureBlock {
+    if (![self.util isAdvertisingTrackingEnabled]) {
+        return;
```
jcazevedo: I think just returning here doesn't give the caller a proper signal that tracking is disabled, particularly because the only interaction callers have with this method is through callbacks. What do you think of calling the `onFailureBlock` with a custom error?

VelocidiSDK/VelocidiSDK/VSDKVelodici.m
```diff
@@ -0,0 +1,63 @@
...
+
+- (void)track:(VSDKTrackingEvent *)trackingEvent {
+    [self track:trackingEvent onSuccess:(void (^)(NSURLResponse *, id)) ^{} onFailure:(void (^)(NSError * error)) ^{}];
+}
+
+- (void)track:(VSDKTrackingEvent *)trackingEvent
+    onSuccess:(void (^)(NSURLResponse *response, id responseObject))onSuccessBlock
+    onFailure:(void (^)(NSError *error))onFailureBlock {
+
+    VSDKRequest * request = [[VSDKRequest alloc] initWithHTTPSessionManager:self.sessionManager];
```
jcazevedo, line 54: Won't this `alloc` call without a corresponding `release` produce memory leaks? I might be overlooking something though. My Objective-C knowledge is very limited.

jcazevedo (changes requested):
I'm sorry for the delay in reviewing this.

We're getting some compiler warnings due to missing nullability type specifiers (we're compiling with `-Wnullability-completeness`) and an incomplete umbrella header. The former is fixable by adding some `nullable` annotations. The latter can be fixed by adding the missing `#import "VSDKRemoveFromCart.h"` and `#import "VSDKSubscription.h"` imports to `VelocidiSDK.h`.

VelocidiSDK/VelocidiSDK/VSDKVelodici.m
duartepinto: No, Objective-C has ARC (https://www.coursera.org/learn/objective-c/lecture/Mn4FV/arc-video), a garbage cleaner. It does not capture everything and there are some edge cases where it does not work, but I think we avoided those cases.

RELEASING.md
duartepinto: No, build number upgrade is automatic (always +1)

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
duartepinto: Changed in 1d3cc3e

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
duartepinto: Changed in 1d3cc3e

RELEASING.md
jcazevedo: I'm getting the following if I run `agvtool new-version -all` without a version:

```
/Users/jcazevedo/work/velocidi-ios-objc-sdk/VelocidiSDK [git::tracking] [jcazevedo@bocuma] [11:53]
> agvtool new-version -all
The new-version operation takes optional -all argument followed by a version number (required).

agvtool - Apple-generic versioning tool for Xcode projects
  usage:
    agvtool help
    agvtool what-version | vers [-terse]
    agvtool [-noscm | -usecvs | -usesvn] next-version | bump [-all]
    agvtool [-noscm | -usecvs | -usesvn] new-version [-all] <versNum>
    agvtool [-noscm | -usecvs | -usesvn] tag [-force | -F] [-noupdatecheck | -Q] [-baseurlfortag]
    agvtool what-marketing-version | mvers [-terse | -terse1]
    agvtool [-noscm | -usecvs | -usesvn] new-marketing-version <versString>
```

Am I missing something?

VelocidiSDK/VelocidiSDK/Tracking/Model/VSDKProduct.h
duartepinto: Removed the _REQUIRED_ in 1d3cc3e. It will rely on the `nonnull` property attribute.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: Changed to @antonio-ramadas implementation in 03c4975

RELEASING.md
duartepinto: Yes, the command is wrong. it should be `agvtool next-version -all` and not `agvtool new-version -all`. Corrected it in 25b1c2b

VelocidiSDK.xcworkspace/xcshareddata/IDETemplateMacros.plist
duartepinto: Yes, this is not needed. I added it in the past but during an offline conversation, we decided not to have any fileheaders. I still kept it just because it looks better to have that being automatically inserted then a `//`. I can remove it if you want

VelocidiSDK/VelocidiSDK/VSDKConfig.h
duartepinto: Added this functionality in 3652ed4

VelocidiSDK/VelocidiSDK/VSDKUtil.m
antonio-ramadas: > I think we can improve this a little bit.
> 
> Why don't we have a private property named `sharedManager` of type `ASIdentifierManager`? When using it, if it is `nil`, then we set it to `[ASIdentifierManager sharedManager]`. This allows us to create a mock class which implements `ASIdentifierManager` and we set it to `sharedManager` (the new property we created).
> 
> I haven't tested this, but it should work. I'm only worried if implementing/extending `ASIdentifierManager` is possible.

@duartepinto What do you think about this?

VelocidiSDK/VelocidiSDK/VSDKRequest.m
duartepinto: Yes, I agree. And this also improves the way we are testing if the request is made. Improved in  e7fd2f6

VelocidiSDK/VelocidiSDK/VSDKUtil.m
duartepinto: Sorry, overlooked this. On it

duartepinto: Implemented this in b8e520a

VelocidiSDK/VelocidiSDK/VSDKRequest.m
```diff
@@ -0,0 +1,71 @@
...
+}
+
+- (void)performRequest: (void (^)(NSURLResponse *response, id responseObject))onSuccessBlock
+        :(void (^)(NSError *error))onFailureBlock {
+    if (![self.identifierManager isAdvertisingTrackingEnabled]) {
+        NSDictionary *userInfo = @{
+                                   NSLocalizedDescriptionKey: NSLocalizedString(@"Operation cannot be completed. Tracking is not allowed", nil),
+                                   NSLocalizedFailureReasonErrorKey: NSLocalizedString(@"The user has opted-out of ad tracking (Limited Ad Tracking is enabled in the user's device)", nil)};
+        NSError * trackingNotAllowedError = [NSError errorWithDomain: VSDKRequest.trackingNotAllowedError code: 1 userInfo: userInfo];
+        return onFailureBlock(trackingNotAllowedError);
```
jcazevedo: This will always be the same error. Can't we reuse it between method calls?

jcazevedo (changes requested):

VelocidiSDK/VelocidiSDK/VSDKRequest.m
duartepinto: Fixed in fb6f6cb

jcazevedo (approved):

antonio-ramadas (approved):

PR #10, opened 2019-06-12
Split README.md

antonio-ramadas:
This PR splits [README.md](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/README.md) into two files: [HEADER.md](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/docs/HEADER.md) and [BODY.md](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/docs/BODY.md).

With this division, we can reference the contents of [README.md](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/README.md) elsewhere without including specifics of the repository (like the coverage or its build status).

I have not found a neat solution to this problem without replicating the files. Unfortunately, I do not see a way for [reStructuredText](https://en.wikipedia.org/wiki/ReStructuredText) helping us. However, I added a new job to Travis so at least we can ensure the files in [docs/](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/docs/) can build [README.md](https://github.com/velocidi/velocidi-ios-objc-sdk/blob/enhancement/split-docs/README.md).


duartepinto (approved):
LGTM! This really seems like the best approach to the problem, well thought 👍 	Small nitpicking, I would prefer the names to be `README_body.md` or `README_BODY.md`, just to make it clear that those files are associated to the README.

antonio-ramadas:
> Small nitpicking, I would prefer the names to be `README_body.md` or `README_BODY.md`, just to make it clear that those files are associated to the README.

It is a good idea. Just to keep the filename consistent I opted for having everything uppercase (fbafe71).

PR #18, opened 2019-07-29
Normalize font family/size according to Velocidi style guide

dimamo5:


beatriz (approved):
LGTM! 

I just noticed a couple of things when testing, which can be tackled separately:
- when building the docs it prints an error on the console `/Users/bmagalhaes/code/velocidi-ios-objc-sdk/VelocidiSDK/VelocidiSDK/VelocidiSDK.h:1:9: fatal error: 'UIKit/UIKit.h' file not found`
- when selecting a property, its name moves left when opening the details:
docs_bug
I'll create issues for them, what do you think?

dimamo5:
It's better. Thank you! 😄

PR #25, opened 2020-07-21
Add AppView event and update Subscription event

beatriz:
This adds a new `appView` event and updates the `subscription` event with duration.


duartepinto (changes requested):
can you add a changlog entry?

beatriz:
> can you add a changlog entry?

Done on 8b404c5d37783d961d39f5b9d90657792c07caab.

duartepinto (approved):
LGTM! The CI is failing when executing `pod install`, but the other tests still pass. Executed the tests that failed locally and they passed. Seems to be something related to [cocoapods using a CDN in the newer versions](https://travis-ci.community/t/xcode-11-2-pod-install-time/5887/8)

PR #27, opened 2020-07-28
Fix compatibility with newer versions of xcode/macos

duartepinto:
This should fix the problem that appeared when running `pod lib lint --verbose VelocidiSDK.podspec` and that was making it impossible to publish a new version in computers with the new xcode/macos versions (unsure which one is the culprit)


beatriz (approved):
LGTM! Maybe it would be nice to have this check on the CI?

duartepinto:
> LGTM! Maybe it would be nice to have this check on the CI?

Yes, that makes sense. I'll open an issue with that

PR #29, opened 2020-08-13
Make JSONModel usage compatible with Swift 3

jcazevedo:
Makes JSONModel usage compatible with Swift 3 by implementing classForCollectionProperty on the class level for collections of nested models.


beatriz (approved):
LGTM, and it's passing the tests on #30 😄

PR #30, opened 2020-08-14
Add tests for the serialization of tracking events

beatriz:
Following the bug that #29 fixes, it looked a good idea to have some tests that check that all events are serializing correctly.

Depends on #29.


duartepinto (approved):
Nice addition 👍

PR #37, opened 2020-10-22
iOS 14 initial support

thyandrecardoso:
Adds basic support to correctly retrieve the IDFA in iOS 14, while maintaining support for previous SDK versions and build tools.


VelocidiSDK/VelocidiSDKTests/UtilTests.swift
```diff
@@ -20,6 +19,24 @@ class UtilTests: QuickSpec {
...
+                        expect(error._domain).to(equal("com.velocidi.VSDKTrackingNotAllowedError"))
+                    })
+                }
+            } else { // previous ios version are opt-out
+                it("should successfuly provide the IDFA") {
+                    if #available(iOS 13, *) {
+                        expect(try? VSDKUtil.tryGetIDFA()).to(contain("00000000-0000-0000-0000-000000000000"))
+                    } else {
+                        expect(try? VSDKUtil.tryGetIDFA()).to(match("[A-Z0-9]{8}-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{12}"))
+                    }
```
duartepinto: I think we can simplify here and use just the regex.
```suggestion
                    expect(try? VSDKUtil.tryGetIDFA()).to(match("[A-Z0-9]{8}-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{12}"))
```

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -61,20 +62,37 @@ class NetworkTests: QuickSpec {
...
+                        success.pointee = false
+                        error.pointee = err
                     })
-
-                    expect(success).toEventuallyNot(beNil(), timeout: 4)
-                    expect(success).toEventuallyNot(beFalse())
+                }
+                
+                if #available(iOS 14, *) { // ios14 is opt-in
+                    it("should fail to execute tracking requests when the user is not opted-in in iOS 14") {
```
jcazevedo, line 80: Can we also test that tracking requests are correctly performed when the user opts-in in iOS 14?

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
```diff
@@ -116,20 +134,37 @@ class NetworkTests: QuickSpec {
...
+                        success.pointee = false
+                        error.pointee = err
                     })
-                    
-                    expect(success).toEventuallyNot(beNil(), timeout: 4)
-                    expect(success).toEventuallyNot(beFalse())
+                }
+                
+                if #available(iOS 14, *) { // ios14 is opt-in
+                    it("should fail to execute match requests when the user is not opted-in in iOS 14") {
```
jcazevedo, line 154: Can we also test that match requests are correctly performed when the user opts-in in iOS 14?

jcazevedo (commented):
I'm not sure we should be disallowing all requests if `tryGetIDFA` fails. I suppose there are valid instances of requests that don't violate the privacy terms if the user doesn't want to be tracked.

Nonetheless, we should probably document somewhere that if the user opts-in (in iOS 14) or opts-out (in iOS <14) from user tracking no requests are ever made, even if they aren't tracking requests and potentially don't include user-specific data.

Examples/SwiftExample/SwiftExample/AppDelegate.swift
```diff
@@ -1,5 +1,6 @@
 import UIKit
 import VelocidiSDK
+import AppTrackingTransparency
```
jcazevedo, line 3: Can we comment somewhere that the example projects must be compiled using Xcode 12?

VelocidiSDK/VelocidiSDK/VSDKUtil.m
```diff
@@ -43,4 +54,42 @@ + (void) setAcceptAllResponses: (AFHTTPSessionManager *)sessionManager {
...
+    if (@available(iOS 14, *)) {
+#if defined(CAN_BUILD_FOR_IOS14)
+        trackingIsAllowed = [ATTrackingManager trackingAuthorizationStatus] == ATTrackingManagerAuthorizationStatusAuthorized;
+#endif
+    } else {
+        trackingIsAllowed = [[ASIdentifierManager sharedManager] isAdvertisingTrackingEnabled];
+    }
+
+    
+    if (trackingIsAllowed) {
```
duartepinto, line 72: I don't think we should be dealing with invalid permissions by receiving an error pointer as function arguments. This pattern isn't really encouraged in obj-c. We should just return nil

thyandrecardoso:
> I'm not sure we should be disallowing all requests if tryGetIDFA fails. I suppose there are valid instances of requests that don't violate the privacy terms if the user doesn't want to be tracked.

> Nonetheless, we should probably document somewhere that if the user opts-in (in iOS 14) or opts-out (in iOS <14) from user tracking no requests are ever made, even if they aren't tracking requests and potentially don't include user-specific data.

I would be OK in changing the way these APIs work, so that we don't implicitly inject the IDFA. We could still provide a helper to use the IDFA but all IDs would be given explicitly. That should address your comment and also make the SDK simpler.

Examples/SwiftExample/SwiftExample/AppDelegate.swift
thyandrecardoso: I put that in the example's `README`: 712566f. 
Is this enough?

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
thyandrecardoso: I don't think we have a way to do that without injecting another dependency in `VSDKVelocidi`.

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
thyandrecardoso: I don't think we have a way to do that without injecting another dependency in `VSDKVelocidi`.

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
thyandrecardoso: Also added 99872d4

VelocidiSDK/VelocidiSDKTests/VelocidiTests.swift
thyandrecardoso: Also added 99872d4

jcazevedo (approved):
Looks good. Thanks!

> I would be OK in changing the way these APIs work, so that we don't implicitly inject the IDFA. We could still provide a helper to use the IDFA but all IDs would be given explicitly. That should address your comment and also make the SDK simpler.

Can you open an issue to keep track of that idea?

thyandrecardoso:
In 4629ee3 I also add some information about changes in iOS 14 and consequences in this SDK. Please take a look.

duartepinto (changes requested):

VelocidiSDK/VelocidiSDK/VSDKUtil.m
thyandrecardoso: I briefly read and tried to follow things like:
- https://stackoverflow.com/questions/53062501/how-to-correctly-handly-nserror-pointers
- https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorHandling/ErrorHandling.html#//apple_ref/doc/uid/TP40001806-CH201-SW1

The whole swift "system" even seems to have automatic signature conversions for this pattern, e.g., `VSDKUtil.tryGetIDFA()`.

duartepinto: Did not know about that. Nevermind then!

VelocidiSDK/VelocidiSDKTests/UtilTests.swift
thyandrecardoso: Sure!

duartepinto (approved):

PR #38, opened 2020-10-26
Remove commited pods

thyandrecardoso:
Clears the `Pods` directories. It should be ok to rely on `pod install` for this, even more now that we have a [CDN for this](https://github.com/velocidi/velocidi-ios-objc-sdk/commit/91a0c9c391fa1250dca0f1e9ed2ff3ed2111346e)


Podfile.lock
```diff
@@ -34,7 +34,7 @@ DEPENDENCIES:
   - Quick (~> 2.0)
 
 SPEC REPOS:
-  https://github.com/cocoapods/specs.git:
+  trunk:
```
jcazevedo, line 37: With CocoaPods 1.10.0, after running `pod install` this gets replaced by `https://cdn.cocoapods.org/:`.

jcazevedo: This seems to have been an issue with my repository configuration for CocoaPods. Nevermind.

jcazevedo (approved):

PR #40, opened 2020-11-09
Docs standardize

duartepinto:
This PR:
* Removes the Guides section in the docs website
* Adds the relevant external links (similar to https://android.developers.velocidi.com)

image

* Updates jazzy from 0.10.0 to 0.13.6

### DIFF
--- a/.jazzy.yaml
+++ b/.jazzy.yaml
@@ -10,15 +10,3 @@ module: VelocidiSDK
 output: ./docs/docs_page/
 theme: docs/velocidi-jazzy-theme/
 readme: ./docs/tmp/readme/README.md
-documentation: 
-  - ./docs/*.md 
-  - ./docs/tmp/docs/*.md 
-custom_categories: 
-  - name: Guides 
-    children: 
-      - Installation
-      - Setting up the SDK
-      - Send a track event
-      - Make a match
-      - Custom tracking events
-      - Tracking model classes list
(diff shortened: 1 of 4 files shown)


.jazzy.yaml
```diff
@@ -10,15 +10,3 @@ module: VelocidiSDK
 output: ./docs/docs_page/
 theme: docs/velocidi-jazzy-theme/
 readme: ./docs/tmp/readme/README.md
```
thyandrecardoso, line 12: Why is this required?

.jazzy.yaml
```diff
@@ -10,15 +10,3 @@ module: VelocidiSDK
...
-documentation: 
-  - ./docs/*.md 
-  - ./docs/tmp/docs/*.md 
-custom_categories:
```
thyandrecardoso, line 16: Do we still need the build steps that split the Readme?

.jazzy.yaml
duartepinto: There has to be some kind of landing page. Thought about removing it, but it would default to just using the whole README as main page.

.jazzy.yaml
thyandrecardoso: Ok, it seems we would still require this script for the `README` part that is used as the main page of the API documentation, and the rest of the partial files are used elsewhere on our documentation building pipelines. So, this is ok.

thyandrecardoso (approved):

PR #41, opened 2020-11-11
Update links to developer documentation in README

jcazevedo:
This PR fixes the location for some links to the developer documentation in the README. It also replaces a link to the examples folder with an absolute URL, so that the README can be easily included elsewhere.


duartepinto (approved):
Nice catch!

PR #43, opened 2021-03-10
Move build to GitHub Actions

thyandrecardoso:
Remove `.travis.yml` and create a new workflow for GitHub Actions.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        include:
+         - os: "macos-10.15"
+           xcode: "10.3"
+           sdk: "12.4"
+           device: "iPhone X"
+         - os: "macos-10.15"
+           xcode: "11.7.0"
+           sdk: "13.7"
+           device: "iPhone 11"
+         - os: "macos-10.15"
```
duartepinto, line 22: Can we add other intermediate sdk versions? According to https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#installed-sdks we should be able to add `13.2`,  `13.3`,  `13.4`,  `13.5`,  `13.6`, `14.0`,  `14.1`, `14.2`, `14.3`

duartepinto (changes requested):

.github/workflows/ci.yml
thyandrecardoso: We would have to change the test structure a bit for that. Also, we probably would not gain that much with those tests, as we should not be relying on things that change between those versions. Given that, I would keep on testing only major versions.

duartepinto: Given what we talked about in an offline conversation, there is really no benefit in us being so thorough in testing this against multiple SDK versions. Other iOS frameworks, like [AFNetworking](https://github.com/AFNetworking/AFNetworking/blob/master/.github/workflows/ci.yml#L29) don't really do that and they are far more complex and prone to use something that would cause errors in earlier versions than us. Testing in versions `12.4`, `13.7`, and `14.4` should be good enough

duartepinto (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+           xcode: "12.4"
+           sdk: "14.4"
+           device: "iPhone 11 Pro Max"
+# NOTICE: Not available for now: https://github.com/actions/virtual-environments/issues/2486
+#         - os: "macos-11.0"
+#           xcode: "12.2"
+#           sdk: "14.0"
+#           device: "iPhone 11 Pro Max"
```
duartepinto: This shouldn't be needed.
```suggestion
```

duartepinto (approved):
LGTM

PR #44, opened 2021-03-12
Change API to stop sending IDFA automatically and ask for all user ids instead

thyandrecardoso:
Changes the API so that the user is required to specify any user id to use on the events.


thyandrecardoso:
@dimamo5 and @duartepinto I synced and resolved conflicts with master, please take a look.

dimamo5 (approved):
LGTM! Can you add an entry to the changelog?

VelocidiSDK/VelocidiSDK/VSDKVelodici.m
```diff
@@ -75,6 +85,22 @@ - (void)match: (NSString *)providerId
...
+    }
+    
+    if ([userIds count] < 2){
+        NSException* ex = [NSException
+                exceptionWithName:@"InvalidArgument"
+                reason:@"At least 2 user ids must be provided!"
+                userInfo:nil];
+        @throw ex;
+    }
+
```
duartepinto: I don't think we should be throwing exceptions here. We should return this error via the `onFailureBlock`

VelocidiSDK/VelocidiSDK/VSDKUserId.m
```diff
@@ -2,10 +2,31 @@
...
+        @throw ex;
+    }
+    
+    if ([userIdType length] <= 0) {
+        NSException* ex = [NSException
+                exceptionWithName:@"InvalidArgument"
+                reason:@"type must not be empty!"
+                userInfo:nil];
+        @throw ex;
+    }
```
duartepinto: I don't think we should be throwing these exceptions here. Perhaps we can do these verifications when we are executing the request and return these errors in the `onFailureBlock`.

VelocidiSDK/VelocidiSDK/VSDKVelodici.m
duartepinto: Also, given that this is internal logic of how the match works, should we be doing this in `VSDKMatchRequest`?

README.md
```diff
@@ -84,16 +84,6 @@ __Objective-C__
...
-__iOS 14 IDFA access__
-
-In iOS 14 Apple introduces changes to the way the Advertising Identifier (IDFA) can be accessed. In particular, developers are forced to ask the user permission to access the IDFA:
-- https://developer.apple.com/app-store/user-privacy-and-data-use/
-- https://developer.apple.com/documentation/apptrackingtransparency?language=objc
-
-The [Examples](https://github.com/velocidi/velocidi-ios-objc-sdk/tree/master/Examples) folder contains two example applications ready for iOS 14, which ask for user permission before using the IDFA to execute any requests.
-
-The Velocidi iOS SDK itself will **not** execute any request if the user did not allow tracking to proceed.
-
```
duartepinto, line 96: We should give more detail on why we don't resort to any kind of internal ID and why it is up to the developer to provide us with the ID it wants. 

We should also document how to fetch the IDFA and/or IDFV, given that those would be the most important ones (I think this will come in a later stage eventually, so we can do this in a different PR)

duartepinto (changes requested):

duartepinto:
I was looking at this, and I think we could simplify our API a bit, to make it easier to use. We could allow the dev to set a default UserId in the config, and we could add an extra method that would use that ID, instead of having to explicitly add it in every request.

README.md
thyandrecardoso: In 175d1db I correct the documentation about match and ask for caution when using the IDFA.

VelocidiSDK/VelocidiSDK/VSDKVelodici.m
thyandrecardoso: In 96af9d1 I try to refactor things to avoid throwing exceptions.

VelocidiSDK/VelocidiSDK/VSDKUserId.m
thyandrecardoso: In 96af9d1 I try to refactor things to avoid throwing exceptions.

thyandrecardoso:
> I was looking at this, and I think we could simplify our API a bit, to make it easier to use. We could allow the dev to set a default UserId in the config, and we could add an extra method that would use that ID, instead of having to explicitly add it in every request.

I think this API change should be left to another PR.

duartepinto (changes requested):
While looking at this, I stumbled upon [Exception Handling and Memory Management](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Tasks/HandlingExceptions.html#//apple_ref/doc/uid/20000059-SW7), which states that Exception handling might complicate memory management in Objective-C. We had an offline discussion and decided that this is something that we want to avoid having to deal with. We talked about reverting these verifications back to `VSDKVelocidi.m` and immediately call the `onFailureBlock` if it fails to pass those assertions.

thyandrecardoso:
> While looking at this, I stumbled upon Exception Handling and Memory Management, which states that Exception handling might complicate memory management in Objective-C. We had an offline discussion and decided that this is something that we want to avoid having to deal with. We talked about reverting these verifications back to VSDKVelocidi.m and immediately call the onFailureBlock if it fails to pass those assertions.

In 8174674 I try to use the callback mechanism. Please take a look.

VelocidiSDK/VelocidiSDK/VSDKUtil.m
```diff
@@ -3,15 +3,6 @@
...
-#define CAN_BUILD_FOR_IOS14 1;
-#endif
-
-#if defined(CAN_BUILD_FOR_IOS14)
-@import AppTrackingTransparency; // NOTICE: linking happens automatically when using Modules or "semantic import".
-#endif
-
 @implementation VSDKUtil
 
 NSString * const trackingNotAllowedErrordomain = @"com.velocidi.VSDKTrackingNotAllowedError";
```
duartepinto: Can you remove this (and the one below)?

duartepinto (changes requested):
LGTM! Just left a small nitpick

duartepinto:
> I think this API change should be left to another PR.

Ok 👍 Can you open an issue with that afterward just so that it doesn't get lost?

VelocidiSDK/VelocidiSDK/VSDKUtil.m
thyandrecardoso: Done in f3edbe5

duartepinto (approved):
LGTM! Thanks for picking this up

dimamo5 (approved):
👍

PR #45, opened 2021-03-13
Minor changes to the project definitions

thyandrecardoso:
We were  targeting `v11` apparently, I think this was unintended.  This tries to correct that and silence warnings over which we cannot do anything at the moment.


duartepinto (changes requested):
LGTM! Can you add an entry to the CHANGELOG stating the change of iOS Deployment target version to 10.0?

thyandrecardoso:
> LGTM! Can you add an entry to the CHANGELOG stating the change of iOS Deployment target version to 10.0?

Yes, I added the changelog entry in 49ba975. Please take a look.

duartepinto (approved):
LGTM! Thanks for this

PR #46, opened 2021-03-14
Add SwiftLint

thyandrecardoso:
This runs [SwiftLint](https://github.com/realm/SwiftLint) as a new "Run Script" phase during compilation of VelocidiSDK tests and SwiftExample. In CI, It also verifies that no diff exists after compilation.

I did not went through the trouble of refactoring the tests to get rid of the method length warning, because we only use Swift for tests. I just ignored it.

The script that runs during the build is the following:
```
"${PODS_ROOT}/SwiftLint/swiftlint" autocorrect && "${PODS_ROOT}/SwiftLint/swiftlint" --strict
```
I am autocorrecting things here. 
I guess we can not do that and I am open to change that behavior. I did it this way, I guess, because the workflow to apply the required changes may be a bit cumbersome given that we may be forced to call the tool by ourselves through the command line... 🤷 . Thoughts?

This depends on #45.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,8 @@ jobs:
           TEST_SDK: "${{ matrix.sdk }}"
           TEST_DEVICE: "${{ matrix.device }}"
         run: ./.github/workflows/xcodebuild.sh clean build test
+      - name: "Verify code linting generated no changes"
+        run: git diff --exit-code
       - name: Upload coverage to Codecov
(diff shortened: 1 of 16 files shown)


thyandrecardoso:
@duartepinto as discussed offline, in 3066e9a and bd1645d I added a helper script to run `swiftlint`. Please take a look.

duartepinto (approved):
LGTM!

PR #47, opened 2021-03-14
Ignore coverage from tests

thyandrecardoso:
This is an attempt to remove coverage from the tests package. 

Using the "ignore" key in the config yml does not seem to be working. Forcing a more specific package seems to work better, although the file SanityTests.m still shows up. I was not able to understand why, nor why the coverage for the tests was not being ignored in the first place using the `ignore: "VelocidiSDK/VelocidiSDKTests/**/*"`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
-          xcode_package: "VelocidiSDK"
+          xcode_package: "^VelocidiSDK$"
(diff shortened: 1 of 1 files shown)


duartepinto:
I think this wasn't happening before. I checked some older commits and they also have that problem, so I would say it is not related to us changing to github actions.
Could be related to this problem
image

If we have a pathing problem, it would explain why the `ignore: "VelocidiSDK/VelocidiSDKTests/**/*"` is not being applied

thyandrecardoso:
Closing because we decided to remove codecov altogether.

PR #48, opened 2021-03-25
Remove domain models

thyandrecardoso:
Removes the tracking events domain models. Refactors API to accept either a JSON string or a NSDictionary.


Examples/SwiftExample/SwiftExample/ViewController.swift
```diff
@@ -56,16 +64,26 @@ class ViewController: UIViewController {
 
     @IBAction func sendCustomTrackingEvent(_ sender: Any) {
```
duartepinto, line 65: I think we can remove the custom examples, given that they are the same as any other event.

duartepinto (changes requested):
Overall looks good to me, although there are some things that I haven't given a thorough review yet.

Can you update the README and CHANGELOG?

thyandrecardoso:
> Can you update the README and CHANGELOG?

Done in 44a8c2f and fcefb7f. Please take a look.

Examples/SwiftExample/SwiftExample/ViewController.swift
thyandrecardoso: I would prefer to leave it. They exhibit a different event type (`custom`) and a different method to use to send a tracking event.

dimamo5 (approved):

.gitattributes
```diff
@@ -1 +1 @@
-*.pbxproj binary
+*.pbxproj binary merge=union
```
duartepinto: Can you leave a comment here explaining why we are adding this?

duartepinto (approved):
LGTM!

.gitattributes
thyandrecardoso: Done in cf5a11a

duartepinto (approved):

duartepinto:
We are aware that codecov is failing! The difference is not relevant and we'll merge this anyway

PR #49, opened 2021-03-25
Add Makefile

duartepinto:
Added a Makefile with the purpose of aggregating commonly used commands. Also added a `prerequisites` script and make command, to make it easier for someone new to bootstrap a new development system. 

I could have added some make commands related to publishing the pod/or documentation, but decided not to, as that is still a manual process and those commands would end up not being really useful on their own.

Also removed the `pod spec lint` from being executed in the tests. This command does not test the current commit, but rather the external repo and associated tag.


.scripts/prerequisites.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+
+if ! which xcodebuild >/dev/null; then
+  echo "xcodebuild does not seem to be installed. You can install it by installing Xcode from the App Store."
+  exit 1
```
thyandrecardoso: It would be more useful if we could exit only when we've run all the tests. We can print all the messages we want, and only exit in the end.

Makefile
```diff
@@ -0,0 +1,44 @@
+XCARGS := -workspace VelocidiSDK.xcworkspace \
```
thyandrecardoso: Please align these properly.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+build:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK build | xcpretty
+
+test:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK test | xcpretty
+
+examples: install-examples build-objc-example build-swift-example
+
+install-examples:
+	set -o pipefail && pod install --project-directory=Examples/
```
thyandrecardoso:
```suggestion
	pod install --project-directory=Examples/
```

Makefile
```diff
@@ -0,0 +1,44 @@
+XCARGS := -workspace VelocidiSDK.xcworkspace \
+		-sdk "iphonesimulator" \
+    -destination "platform=iOS Simulator,OS=14.4,name=iPhone 11 Pro Max" \
+    GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES GCC_GENERATE_TEST_COVERAGE_FILES=YES CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
+
+all: clean build
+
+clean:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK  clean | xcpretty
```
thyandrecardoso: Shouldn't this clean the examples too?

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	set -o pipefail && \
+	xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace clean && \
+	xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO | xcpretty -r json-compilation-database --output compile_commands.json && \
+	oclint-json-compilation-database -exclude Pods -exclude build -- -report-type xcode -max-priority-3=15000
+
+swiftlint:
+	set -o pipefail && Pods/SwiftLint/swiftlint lint --fix && Pods/SwiftLint/swiftlint lint --strict
+
+podlint:
+	set -o pipefail && pod lib lint --verbose
```
thyandrecardoso:
```suggestion
	pod lib lint --verbose
```

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	.scripts/prerequisites.sh
+
+oclint:
+	set -o pipefail && \
+	xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace clean && \
+	xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO | xcpretty -r json-compilation-database --output compile_commands.json && \
+	oclint-json-compilation-database -exclude Pods -exclude build -- -report-type xcode -max-priority-3=15000
+
+swiftlint:
+	set -o pipefail && Pods/SwiftLint/swiftlint lint --fix && Pods/SwiftLint/swiftlint lint --strict
```
thyandrecardoso:
```suggestion
	Pods/SwiftLint/swiftlint lint --fix && Pods/SwiftLint/swiftlint lint --strict
```

.github/workflows/ci.yml
```diff
@@ -47,10 +47,6 @@ jobs:
     name: "Build example applications"
```
thyandrecardoso, line 49: It feels a bit strange to have all the build commands inside the `Makefile` and then the still use the `./.github/workflows/xcodebuild.sh` file in the tests.

Can't we use the `make` there too? I think it is possible to import env vars into `Makefile`s.

.scripts/prerequisites.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
```
thyandrecardoso, line 1: Please add this file to the xcode group, so that it is visible inside the project.

Makefile
thyandrecardoso: Please add this file to the xcode group, so that it is visible inside the project.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+install:
+	pod install
+
+prerequisites:
+	.scripts/prerequisites.sh
+
+oclint:
+	set -o pipefail && \
+	xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace clean && \
```
thyandrecardoso: Can't we save a line here and do `xcodebuild -scheme VelocidiSDK -sdk iphoneos -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO clean build` instead?

Makefile
```diff
@@ -0,0 +1,44 @@
...
+build-swift-example:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme SwiftExample clean build | xcpretty
+
+install:
+	pod install
+
+prerequisites:
+	.scripts/prerequisites.sh
+
+oclint:
```
thyandrecardoso, line 44: I've had some challenges running this.

```
pyenv: python3: command not found

The `python3' command exists in these Python versions:
  3.7.9

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
make: *** [oclint] Error 127
```

```
dlopen(/usr/local/bin/../lib/oclint/rules/libTooManyMethodsRule.dylib, 1): no suitable image found.  Did find:
	/usr/local/bin/../lib/oclint/rules/libTooManyMethodsRule.dylib: code signature in (/usr/local/bin/../lib/oclint/rules/libTooManyMethodsRule.dylib) not valid for use in process using Library Validation: library load disallowed by system policy

oclint: error: cannot open dynamic library: /usr/local/bin/../lib/oclint/rules/libTooManyMethodsRule.dylib
make: *** [oclint] Error 1
```

Screenshot 2021-03-26 at 12 14 59

thyandrecardoso, line 44: Shouldn't this oclint run also for the examples?

.scripts/prerequisites.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "xcodebuild does not seem to be installed. You can install it by installing Xcode from the App Store."
+  exit 1
+fi
+
+if ! which pod >/dev/null; then
+  echo "Cocoapods does not seem to be installed. You may need to run 'gem install cocoapods'."
+  exit 1
+fi
+
+if ! which xcpretty >/dev/null; then
```
thyandrecardoso, line 15: These tools are all mostly dead :(
The project seems mostly dead: https://github.com/xcpretty/xcpretty/issues/360

thyandrecardoso (changes requested):

Makefile
duartepinto: Fixed in 6aaf770

Makefile
duartepinto: Done in 6aaf770

.scripts/prerequisites.sh
duartepinto: Done in 323259f

Makefile
duartepinto: Done in 323259f

.scripts/prerequisites.sh
duartepinto: Done in a698e94

Makefile
duartepinto: Done in 3d01567

Makefile
duartepinto: I don't think I ever had this problem. I tried uninstalling and installing oclint and I could not replicate this issue. I don't use `pyenv`, so it might be related to that.

thyandrecardoso: Regardless of `pyenv`, I think `python3` is required. 

The second problem is not related to python.

duartepinto: Added a command for this in 3025b2c

.github/workflows/ci.yml
duartepinto: Done in 9160da5

Makefile
```diff
@@ -0,0 +1,55 @@
+WORKSPACE ?= VelocidiSDK.xcworkspace
+TEST_SDK ?= 14.4
+TEST_DEVICE ?= iPhone 11 Pro Max
+
+XCARGS := -workspace $(WORKSPACE) \
+					-sdk "iphonesimulator$(TEST_SDK)" \
+					-destination "platform=iOS Simulator,OS=$(TEST_SDK),name=$(TEST_DEVICE)" \
+					GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES GCC_GENERATE_TEST_COVERAGE_FILES=YES CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
+
+all: clean build
```
thyandrecardoso: Nitpick: I think that it is most common to have the default target be only the build phase. This allows one to take advantage of an incremental compilation while repeatedly calling `make`. 

`clean` is normally called separately.

Makefile
```diff
@@ -0,0 +1,55 @@
...
+# we have to clean schemas independently because xcode does not allow to clean all schemes in a workspace
+clean:
+	xcodebuild $(XCARGS) -scheme VelocidiSDK clean | xcpretty && \
+	xcodebuild $(XCARGS) -scheme ObjcExample clean | xcpretty && \
+	xcodebuild $(XCARGS) -scheme SwiftExample clean | xcpretty
+
+build:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK build | xcpretty
+
+test:
```
thyandrecardoso: Shouldn't the `test` target depend on the `build` target?

Makefile
```diff
@@ -0,0 +1,55 @@
...
+	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK test | xcpretty
+
+examples: install-examples build-objc-example build-swift-example
+
+install-examples: all
+	pod install --project-directory=Examples/
+
+build-objc-example:
+	set -o pipefail && xcodebuild $(XCARGS) -scheme ObjcExample clean build | xcpretty
```
thyandrecardoso, line 28: Shouldn't these examples depend on the `install-examples` target?

thyandrecardoso (changes requested):

Makefile
duartepinto: Yes, that makes more sense! Fixed in 173f363

Makefile
duartepinto: Yes. Fixed in 173f363

Makefile
duartepinto: Fixed in 173f363

thyandrecardoso (approved):
Looks fine.

In the future, I think we should try to actually force the use of `oclint` which is, right now, an optional step of the build. Unlike `swiftlint` that runs as part of the compilation step, we don't force `oclint` to run anywhere. It would be nice if we could make a better use of it even automate its recommendations.

We also have some other steps that might be integrated with the `Makefile`, such as the documentation build and publishing steps.

PR #52, opened 2021-04-01
Add OClint CI step and clang-format

duartepinto:
Cleaned up some P3 warnings (and silenced some others) and added OClint as a step in CI. To facilitate this, I also added a formatter, `clang-format`. The formater will automatically run everytime the `VelocidiSDK` and `ObjcExample` schemes are built.

P3 warnings corrected were mostly line length (with `clang-format`) and unused variables. The unused variables were due to having variables in the anonymous functions for `onFailure` and `onSuccess` callbacks. At the time I didn't removed them because I was unsure if it would make them those functions too confusing  to interpret, due to objc syntax, but I now don't really think that is reason enough to keep them.

Breaking long lines makes it necessary to add selectors to each parameter, so I had to add those. Ex: VSDKConfig initializer became: 
```objc
- (instancetype)initWithTrackingBaseUrl:(NSString *)trackingBaseUrl
                           matchBaseUrl:(NSString *)matchBaseUrl;
```

Also added a new command to the Makefile, `test-all`, just for convenience.


Examples/ObjcExample/ObjcExample/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -92,7 +92,6 @@
     }
```
thyandrecardoso: Nitpick: can we avoid this change?

Examples/ObjcExample/ObjcExample/Assets.xcassets/Contents.json
```diff
@@ -1,6 +1,3 @@
 {
```
thyandrecardoso: Nitpick: can we avoid this change?

Examples/ObjcExample/ObjcExample/ViewController.m
```diff
@@ -10,101 +10,135 @@ @interface ViewController ()
...
+      \"type\": \"custom\",\
+      \"customFields\": {\
+        \"key\": \"value\"\
+      },\
+      \"customType\": \"custom-type\"\
+    }\
+    ";
+    // clang-format on
+
+    __attribute__((annotate(
```
thyandrecardoso: This seems way too intrusive.

I would rather just rename the variable.

Makefile
```diff
@@ -10,6 +10,11 @@ XCARGS := -workspace $(WORKSPACE) \
 build:
 	set -o pipefail && xcodebuild $(XCARGS) -scheme VelocidiSDK build | xcpretty
 
+format:
+	set -o pipefail && \
```
thyandrecardoso: I don't think we need this here because we are not using pipes.

Makefile
```diff
@@ -38,16 +43,19 @@ prerequisites:
 
 oclint-examples:
 	set -o pipefail && \
+	xcodebuild -scheme VelocidiSDK -sdk iphonesimulator -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO clean build && \
 	xcodebuild -scheme ObjcExample -sdk iphonesimulator -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO clean build | xcpretty -r json-compilation-database --output compile_commands.json && \
-	oclint-json-compilation-database -exclude Pods -exclude build -- -report-type xcode -max-priority-3=15000
+	oclint-json-compilation-database -exclude Pods -exclude build -- -report-type xcode
```
thyandrecardoso, line 47: Does this command exit with error if there are warnings or errors in the report?

thyandrecardoso (changes requested):
`clang-format` seems to be using `ColumnLimit:     80`. However, that seems a rather tight column limit.

Should we consider increasing that to a limit more commonly found in out there? Or something in line with our internal tools and projects (120).

In Android, we use 100. That is also the line length followed in google style guide: https://google.github.io/styleguide/objcguide.html#line-length

Apple has its own style guide for Cocoa: https://google.github.io/styleguide/objcguide.html#line-length Haven't found anything about line length though.

Makefile
duartepinto: Yes

Makefile
duartepinto: Removed in 03ce0d1

Examples/ObjcExample/ObjcExample/Assets.xcassets/Contents.json
duartepinto: Reverted in 5edb8df

Examples/ObjcExample/ObjcExample/Assets.xcassets/AppIcon.appiconset/Contents.json
duartepinto: Reverted in 5edb8df

Examples/ObjcExample/ObjcExample/ViewController.m
duartepinto: Renamed in 431afba

duartepinto:
> `clang-format` seems to be using `ColumnLimit: 80`. However, that seems a rather tight column limit.
> 
> Should we consider increasing that to a limit more commonly found in out there? Or something in line with our internal tools and projects (120).
> 
> In Android, we use 100. That is also the line length followed in google style guide: [google.github.io/styleguide/objcguide.html#line-length](https://google.github.io/styleguide/objcguide.html#line-length)
> 
> Apple has its own style guide for Cocoa: [google.github.io/styleguide/objcguide.html#line-length](https://google.github.io/styleguide/objcguide.html#line-length) Haven't found anything about line length though.

Changed to 100 in 0fba96b

thyandrecardoso (approved):

PR #53, opened 2021-04-14
Update how IDFA is accessed in Examples

duartepinto:
This PR changes how the IDFA is accessed in the Example projects. Instead of requesting it once and storing it, we always request it before using it. The seems to be more in line with how it is intended to be used and avoids situations where the privacy settings changed but we keep using the the IDFA.

### DIFF
--- a/Examples/ObjcExample/ObjcExample/ViewController.m
+++ b/Examples/ObjcExample/ObjcExample/ViewController.m
@@ -12,21 +12,26 @@
 - (void)viewDidLoad {
   [super viewDidLoad];
   // Do any additional setup after loading the view, typically from a nib.
+}
 
-  self.trackingIsAllowed = false;
++ (void)useIDFA:(void (^)(bool, NSString *))completionHandler {
   if (@available(iOS 14, *)) {
     [ATTrackingManager requestTrackingAuthorizationWithCompletionHandler:^(
                            ATTrackingManagerAuthorizationStatus status) {
-      self.trackingIsAllowed = status == ATTrackingManagerAuthorizationStatusAuthorized;
-      if (self.trackingIsAllowed) {
-        self.idfa = [[[ASIdentifierManager sharedManager] advertisingIdentifier] UUIDString];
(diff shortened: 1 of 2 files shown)


duartepinto:
I'm aware of the CI step failing and will be looking into it separately.

Examples/ObjcExample/ObjcExample/ViewController.m
```diff
@@ -12,117 +12,128 @@ @implementation ViewController
...
                            ATTrackingManagerAuthorizationStatus status) {
-      self.trackingIsAllowed = status == ATTrackingManagerAuthorizationStatusAuthorized;
-      if (self.trackingIsAllowed) {
-        self.idfa = [[[ASIdentifierManager sharedManager] advertisingIdentifier] UUIDString];
+      bool isTrackingEnabled = status == ATTrackingManagerAuthorizationStatusAuthorized;
+      NSString *idfa = nil;
+      if (isTrackingEnabled) {
+        idfa = [[[ASIdentifierManager sharedManager] advertisingIdentifier] UUIDString];
       }
+      completionHandler(isTrackingEnabled, idfa);
```
thyandrecardoso, line 26: Can we simplify this and avoid calling the hook when the IDFA is not available? That would avoid those `if (isTrackingEnabled)` on each method, and I think we can fill the label in right here instead, in this method.

thyandrecardoso (changes requested):

Examples/ObjcExample/ObjcExample/ViewController.m
duartepinto: I would prefer not to, since this serves as example, it's preferable to be as generic as possible. This function could be in a separate class where it would not have access to the label

thyandrecardoso: I am ok with that.

However, I think that if we are looking at these as something that everyone will use as is, we should probably provide it as part of the SDK (I am not suggesting that).

thyandrecardoso (approved):

PR #54, opened 2021-04-15
Fix CI

duartepinto:
* Removes tests for xcode 10.3/sdk 12.4. Running `pod install` resulted in: 
```
Couldn't determine repo type for URL: `https://cdn.cocoapods.org/`: Compilation error generating constants :
	ld: unsupported tapi file type '!tapi-tbd' in YAML file 
```
Not really sure why this started failing without us introducing any change, but it only seems to be failing in that version.
* All `make` commands use the same `sdk` and `destination`. This avoids a mismatch with including dependencies that were built for a different target and will fix the Oclint step not working.

### DIFF
--- a/Examples/ObjcExample/ObjcExample.xcodeproj/project.pbxproj
+++ b/Examples/ObjcExample/ObjcExample.xcodeproj/project.pbxproj
@@ -108,7 +108,7 @@
 				30BCF73622709C1E00E462A8 /* Frameworks */,
 				30BCF73722709C1E00E462A8 /* Resources */,
 				4E82CBF95466188B0C93253D /* [CP] Embed Pods Frameworks */,
-				30F36CAC261627560025FCBF /* Run Script */,
+				30F36CAC261627560025FCBF /* Run clang-format */,
 			);
 			buildRules = (
 			);
@@ -165,7 +165,7 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
-		30F36CAC261627560025FCBF /* Run Script */ = {
+		30F36CAC261627560025FCBF /* Run clang-format */ = {
 			isa = PBXShellScriptBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
@@ -174,14 +174,14 @@
 			);
 			inputPaths = (
 			);
-			name = "Run Script";
+			name = "Run clang-format";
 			outputFileListPaths = (
 			);
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "source ~/.bash_profile\ncd ${SRCROOT}\ncd ../..\nmake install-examples\n\nclang-format -style=file -i Examples/ObjcExample/**/*.m Examples/ObjcExample/**/*.h\n";
+			shellScript = "source ~/.bash_profile\ncd ${SRCROOT}\ncd ../..\nclang-format -style=file -i Examples/ObjcExample/**/*.m Examples/ObjcExample/**/*.h\n";
 		};
 		4E82CBF95466188B0C93253D /* [CP] Embed Pods Frameworks */ = {
 			isa = PBXShellScriptBuildPhase;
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -42,13 +42,12 @@ prerequisites:
 
 oclint-examples:
 	set -o pipefail && \
-	xcodebuild -scheme VelocidiSDK -sdk iphonesimulator -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO clean build && \
-	xcodebuild -scheme ObjcExample -sdk iphonesimulator -workspace VelocidiSDK.xcworkspace COMPILER_INDEX_STORE_ENABLE=NO clean build | xcpretty -r json-compilation-database --output compile_commands.json && \
+	xcodebuild -scheme ObjcExample $(XCARGS) COMPILER_INDEX_STORE_ENABLE=NO clean build | xcpretty -r json-compilation-database --output compile_commands.json && \
```
thyandrecardoso, line 45: How does this work? Shouldn't this depend on the SDK itself so that the example can be built?

thyandrecardoso (changes requested):

Makefile
duartepinto: `VelocidiSDK` is installed as a Pod of `ObjcExample` when we make `pod install --project-directory=Examples`, and that's the version that will be used. That's why even though we were compiling `VelocidiSDK` here, we were still getting errors with linking to a class that was built for a different target.

duartepinto:
In e46f233 I've changed the Makefile and Xcode `Run Scripts` project configurations to always run `make install-examples` before compiling the Example projects

Examples/ObjcExample/ObjcExample.xcodeproj/project.pbxproj
```diff
@@ -181,7 +181,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "source ~/.bash_profile\ncd ${SRCROOT}\ncd ..\nclang-format -style=file -i ObjcExample/**/*.m ObjcExample/**/*.h\n";
+			shellScript = "source ~/.bash_profile\ncd ${SRCROOT}\ncd ../..\nmake install-examples\n\nclang-format -style=file -i Examples/ObjcExample/**/*.m Examples/ObjcExample/**/*.h\n";
```
thyandrecardoso: This script does not seem to prevent us from manually running `pod install --project-directory=Examples` before trying to compile. You can seen the problem when trying to run this from a clean state.

thyandrecardoso (changes requested):

Examples/ObjcExample/ObjcExample.xcodeproj/project.pbxproj
duartepinto: Yes, it does not alleviate the need for a first `pod install --project-directory=Examples`. I could not get that work, but I didn't really delve deep into it. This will automatically install VelocidiSDK if any changes were made to it after that.

thyandrecardoso: Ok. But then this is kinda misleading: the `install` part does nothing for the current compilation session. 

Wouldn't we be better off without `make install-examples` in this script?

duartepinto: > Ok. But then this is kinda misleading: the `install` part does nothing for the current compilation session.

Yes, that is true. I could not find a way to add a script that would run before the `Check Dependencies` phase, that is why this fails to avoid that.

> Wouldn't we be better off without `make install-examples` in this script?

I find it useful to have any change to VelocidiSDK be automatically added when running one of the examples project, hence why I added this. 

I can add a comment stating that `make install-examples` is just being executed to automatically include any change made in VelocidiSDK.

thyandrecardoso: > I find it useful to have any change to VelocidiSDK be automatically added when running one of the examples project, hence why I added this.

Sure, but doesn't that usefulness come from having the example picking up the most recent changes?

From what I understand, with the current modifications on this PR, the version of the `VelocidiSDK` that is installed when you compile one of the examples won't be the version that the example is running against. So, you might be running an example expecting to see the behavior of some change you made inside `VelocidiSDK` but that won't happen... or am I missing something?

duartepinto: You're right, this wasn't working as intended. I tested this by changing the content of the User-Agent sent, and it worked, but stuff that is verified on compilation (like linking methods), will not work and will throw an error. Removed in 5b154e4

thyandrecardoso (approved):

PR #55, opened 2021-04-15
Update README with information on how to manage user ids

duartepinto:
Depends on https://github.com/velocidi/velocidi-ios-objc-sdk/pull/54


README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
     }];
 ```
 
-If you are using the IDFA to track your user, please beware that since iOS 14 Apple is more strict on the IDFA use cases. Please check Apple [documentation on which conditions](https://developer.apple.com/app-store/user-privacy-and-data-use/) and [how to retrieve the user's IDFA](https://developer.apple.com/documentation/apptrackingtransparency?language=objc).
+# iOS 14 and collecting IDs
+
+Due to changes introduced in iOS 14, VelocidiSDK no longer uses the IDFA by default and requires the developer to explicit an ID to identify the user. It is up to the developer to choose which user IDs to use. As a rule of thumb, we recommend reading Apple's
+instructions on [User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/).
+
```
thyandrecardoso: I think I would prefer a preamble where we immediately refer to Apple documentation:

```suggestion

In iOS 14, Apple changed their privacy guidelines and APIs. We recommend reading Apple's
instructions on [User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/).

Due to those changes the VelocidiSDK no longer uses the IDFA by default and instead requires the developer to explicitly define an ID to 
identify the user. It is up to the developer to choose which user IDs to use, taking in consideration that the ID type should be supported by 
the CDP system. Refer to https://docs.velocidi.com/collect/user-ids/#default-id-types for currently supported IDs.

```

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+## Using your own first-party ID
+You can use any first-party ID with [VSDKUserId](https://ios.developers.velocidi.com/Classes/VSDKUtil.html), as exemplified in
```
thyandrecardoso: I don't think this is entirely true.

The ID type and format must be supported and correctly configured in the CDP instance.

I think we should indicate just that, and also point to the supported id types configuration.

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+[Make a Match](#Make-a-match) and [Send a tracking event](#Send-a-tracking-event).
+
+## Using the IDFV
+The [Identifier for Vendor (IDFV)](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) is an ID that is shared amongst apps from
```
thyandrecardoso, line 243: According to Apple:

```
The IDFV may not be combined with other data to track a user across apps and websites owned by other companies unless you have been granted permission to track by the user.
```

Because of this, I am unsure up until where we should be explaining or suggesting people to use `idfv` or `idfa`.

But, assuming we want to carry on this direction, I would be as cautious as we are in the `Using the IDFA`. I would probably quote/copy Apple's description here, and also point to https://developer.apple.com/app-store/user-privacy-and-data-use/.

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+```objectivec
+NSString *idfv = [[[UIDevice currentDevice] identifierForVendor] UUIDString];
+VSDKUserId *userId = [[VSDKUserId alloc] initWithId:idfv type:@"idfv"];
+```
+
+## Using the IDFA
+If you are using the Identifier for Advertisers (IDFA) to identify the user, please make sure to read Apple's instructions on
+[User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/) and be sure that your use case is compliant with Apple's guidelines.
+
+Using the IDFA requires explicit authorization from the user for each application. To do so, start by establishing the message that will be presented to the user in the permission dialog. This can be done by adding the __NSUserTrackingUsageDescription__ key to the Info.plist file of the application. The value of this key will be the message presented.
```
thyandrecardoso, line 265: Can't we just point to https://developer.apple.com/documentation/apptrackingtransparency?language=objc instead of explaining details such as the `Info.plist` modification?

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+
+Before starting to send events, we have to request permission to access the IDFA, and only if that permission is granted, can we send events with
+the IDFA.
+
+__Swift__
+```swift
+import AppTrackingTransparency
+import AdSupport
+
+func useIDFA(completionHandler: @escaping (Bool, String) -> Void) {
```
thyandrecardoso: Why is this `@escaping` necessary?

thyandrecardoso: Why do we need to provide a `Bool` here? Can't we call `completionHandler` only when tracking is permitted, and with a single argument describing the IDFA value? The value could be `nil` if tracking is not allowed...?

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+
+## Using the IDFA
+If you are using the Identifier for Advertisers (IDFA) to identify the user, please make sure to read Apple's instructions on
+[User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/) and be sure that your use case is compliant with Apple's guidelines.
+
+Using the IDFA requires explicit authorization from the user for each application. To do so, start by establishing the message that will be presented to the user in the permission dialog. This can be done by adding the __NSUserTrackingUsageDescription__ key to the Info.plist file of the application. The value of this key will be the message presented.
+
+![tracking usage description entry](./docs/ios14_prompt_message.png)
+
+Before starting to send events, we have to request permission to access the IDFA, and only if that permission is granted, can we send events with
```
thyandrecardoso: I think we should more stronly present the following code excerpts as examples, stating that it is the developers decision when, where and how to present the permission dialog to the user, and the following are just utility methods that can be used if one does not care to control the moment when a permission dialog is shown to the user...

README.md
```diff
@@ -224,10 +224,122 @@ __Objective-C__
...
+```objectivec
+[self useIDFA:^(bool isTrackingEnabled, NSString *idfa) {
+  if (isTrackingEnabled) {
+    VSDKUserId *userId = [[VSDKUserId alloc] initWithId:idfa type:@"idfa"];
+    // ...
+  }
+}];
+```
+
+More information on how to retrieve the user's IDFA can be found in Apple's documentation of the [App Tracking Transparency framework](https://developer.apple.com/documentation/apptrackingtransparency?language=objc).
```
thyandrecardoso: I think the pointer to the App Tracking Transparency framework should be put at the beginning of this section.

thyandrecardoso (changes requested):

README.md
duartepinto: To use it inside the `completionHandler` of `ATTrackingManager.requestTrackingAuthorization`

duartepinto: Yes, that is indeed simpler. Changed in 6da3600

duartepinto:
Given that @thyandrecardoso is on vacation, I've asked @dimamo5 to review this.

README.md
duartepinto: Done in ce9fb2d

README.md
duartepinto: Done in ce9fb2d

README.md
duartepinto: Changed in 6804905

README.md
duartepinto: Changed in 6804905

README.md
duartepinto: The issue to revise this explicitly states that we should provide examples on how to fetch these IDs, so I included all the steps to avoid any confusion.

dimamo5 (approved):

PR #57, opened 2022-08-09
Update README with Kevel information

duartepinto:
While doing this I also wrapped lines bigger than 120 characters and fixed some links


beatriz (approved):
Thanks for this!

PR #58, opened 2022-08-10
Fix build CI

duartepinto:
* Added `Podfile.lock` and `Examples/Podfile.lock` as excluded from the `Verify code linting generated no changes` step
* Froze the macOS version to `10.15` in `build-example-applications` and `run-oclint` jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       - name: "Build example Swift application"
         run: make build-swift-example
       - name: "Verify code linting generated no changes"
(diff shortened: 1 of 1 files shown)


thyandrecardoso (approved):
This seems fine for now. 

It is confusing to me why there are _lockfiles_ that change each time a `install` is issued. https://github.com/CocoaPods/CocoaPods/issues/8848 the answers here are not satisfactory...

We should investigate this further, or change our processes to guarantee that, for example, publishing this repo is done using the expected cocopods version, given that it seems to be a problem worth tracking inside the lockfile.

PR #61, opened 2024-05-16
Update GitHub actions versions

xRuiAlves:
Updating GitHub actions to the latest released stable version.

This PR was programmatically created by the `github-actions-version-updater` script.

Note that further manual action may be required if the action version updating introduces breaking changes.

### Testing

Relying on CI. Please validate the changes for all of the updated GitHub Actions workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,11 @@ jobs:
     strategy:
       matrix:
         include:
-         - os: "macos-10.15"
+         - os: "macos-latest"
            xcode: "11.7.0"
            sdk: "13.7"
            device: "iPhone 11"
-         - os: "macos-10.15"
+         - os: "macos-latest"
            xcode: "12.4"
            sdk: "14.4"
            device: "iPhone 11 Pro Max"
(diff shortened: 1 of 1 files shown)


xRuiAlves:
@jcazevedo MacOs/iOs version 10 instances are already deprecated:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

To use a newer instances, we would also need to change the platform we are targetting in our podfile and podspec, and that needs further testing and work ([different SDK / xcode versions need to be configured as well for more recent macOs/iOs versions](https://developer.apple.com/support/xcode/)) that fall outside the scope of this PR (and which I really don't know how to navigate lol, never done anything iOs related).

That being said, I propose we either force push this (as CI won't run because there isn't a CI instance type that will pick these jobs), or we close it altogether. Let me know what you think?

jcazevedo:
> That being said, I propose we either force push this (as CI won't run because there isn't a CI instance type that will pick these jobs), or we close it altogether. Let me know what you think?

I'm fine with force-pushing, but, given that the only instructions on how to build this seem to be the CI workflow, I think we should get that fixed as soon as possible.

jcazevedo (approved):

